PR #6, opened 2015-11-10
ported to Python 2

spacediver:
I'd request for more profound tests, maybe, but basically, external crypto libs are called properly, tokens are created and deleted.

### DIFF
--- a/knox/models.py
+++ b/knox/models.py
@@ -16,7 +16,7 @@ class AuthTokenManager(models.Manager):
         if expires is not None:
              expires = timezone.now() + expires
 
-        auth_token = super().create(digest=digest, salt=salt, user=user, expires=expires)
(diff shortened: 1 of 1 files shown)


James1345:
Merged your python2 changes across to the develop branch. I'd advise you use that for any more work you want merged in, master is reserved for releases (and merging across branches is not the nicest task)
I've seen your change to add create_unsafe, Out of interest what is that for?

spacediver:
In my application (REST API with AuthToken header in each request) I need to know, at the moment of authentication, from which device this user is coming. I do not want to force specifying device_id in each request, instead, I force it only on login and signup, when the very tokens are generated. Then, token_id is linked to device, so that I can tell the device simply by looking up `request.auth`, which is AuthToken instance.

PR #12, opened 2016-02-12
Separate knox's default authentication for login from the DRF's one 

Corea:
To use knox's TokenAuthentication for DRF's default authentication, we need to separate knox's default authentication from the DRF's one.


James1345:
Merged into develop branch

PR #14, opened 2016-04-02
Token duration and use count as parameters

dryvenn:
Hi @James1345 ,
First of all, thank you for open-sourcing this app, it helps a lot!

I had specific needs for my project, so I added some features:
- the ability to specify the token validity time as a query parameter;
- token validity based on use count rather than time.

I don't know if this corresponds to the philosophy of the project, but if you want me to split the PR or to make some modifications, I'd be happy to do it.


James1345:
I have no problem with these features. I shall merge them in as options alongside the others next time I have the time to do a release of this.

Thanks :)

James1345:
Hey, I know it's been a while that this has been open. Just re-addressing all of the pull requests.

I have a couple of concerns with this, upon review. Nothing major, but should be addressed before it can be pulled.

Is there a reason you remove the expiry and use a time+creation field? I would be hesitant to add this, as creation time is more information that has the possibility of aiding an attacker. Could you not calculate the expiry date from the two parameters at model creation (while still having the option to set short-dated tokens in the query url). (on that note, the migration between models would remove information in systems where this is already deployed. To merge this the migration must preserve the functionality, possibly set creation=0 and time=expiry before deleting the expiry field)

dryvenn:
Hi,
Yeah it's been a while so I don't recall the exact reason. But looking at the code I guess I found the `expires` field to be redundant with the `created` field, which was already there. I guess you added it as a reference for admins.

OK for the migration matter.

dryvenn:
This should do the job regarding migration. Let me know what you think about this `created` field and I'll do the changes if necessary.

belugame:
@dryvenn To be honest the idea of indefinitely valid tokens sounds like an unusual requirement. Can you elaborate about the usecase?

dryvenn:
Hi,
This PR is not about infinite tokens, but rather about:
- choosing dynamically the duration of a token, using a parameter given by the client (and possibly capped by a setting);
- having token that are not time-based, but usage-based, meaning that they can only be used a given number of time.

Tokens would therefore be based on a time expiry, a usage expiry, or both, which doesn't qualify as infinite.

belugame:
Since this has not seen support or demand from other rest-knox users and is also outdated by now I'm going to close it.

PR #16, opened 2016-04-08
Don't assume User has an "username" field

ko-pp:
Custom User models might not have an "username" field but they declare a USERNAME_FIELD variable which indicate which field to use instead.

See Django official example: https://docs.djangoproject.com/en/1.9/topics/auth/customizing/#a-full-example

In any case, thank you for this great app :-)

### DIFF
--- a/knox/serializers.py
+++ b/knox/serializers.py
@@ -7,4 +7,4 @@ User = get_user_model()
 class UserSerializer(serializers.ModelSerializer):
     class Meta:
         model = User
-        fields = ('username', 'first_name', 'last_name',)
+        fields = (User.USERNAME_FIELD, 'first_name', 'last_name',)


ko-pp:
In fact, it's probably as bad an idea to forcibly try to send "first_name" and "last_name", they aren't mandatory fields either. It would probably be better to have the dev create a /me endpoint in their API to get information about the user, or perhaps create a variable in the settings listing the fields they want to get when /auth/login is called.

James1345:
Some comments:
The USERNAME_FIELD setting was new in Django 1.8. I can't find my original list of supported versions, but I was fairly sure I supported Django 1.7 at the last release, so this field was not valid for use. I shall reconsider it at the next release.

The knox setting for a custom User Serializer already provides functionality for custom user models. The provided serializer is a basic one that assumes the django default User Model is used. The documentation will be updated to indicate this more clearly.

PR #19, opened 2016-07-14
Added token name & customizable auth header prefix

tarequeh:
- Adding token names to meaningfully separate tokens
- Made auth header prefix customizable


tarequeh:
Sorry meant to create this on the forked repo.

PR #24, opened 2016-10-24
Fixed UserSerializer, so it accounts USERNAME_FIELD

davydany:
When using knox, if I have a custom model that doesn't use username as the username field, like say email_address, the UserSerializer fails because Django lets us have custom username fields. (See: https://docs.djangoproject.com/en/1.10/topics/auth/customizing/#django.contrib.auth.models.CustomUser.USERNAME_FIELD)

### DIFF
--- a/knox/serializers.py
+++ b/knox/serializers.py
@@ -4,7 +4,9 @@ from rest_framework import serializers
 
 User = get_user_model()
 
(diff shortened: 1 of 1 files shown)


belugame:
see 8fc7b0f3f9b30ef70ce738936027de6745cfab85

davydany:
@belugame  I'm confused. You said "see `8fc7b0f`", but there is no comment there, and that is the fix that I introduced. Are you closing this ticket because you don't like the fix?

PR #26, opened 2016-11-07
Move UserSerializer definition into class

belugame:
To avoid error when extending LoginView: django.core.exceptions.AppRegistryNotReady: Models aren't loaded yet
@James1345  Hope you find time to have a look at this :)


belugame:
see https://github.com/James1345/django-rest-knox/commit/15ae24d4f07dc142ce828d9ea93271ce02f0ba83

Issue #21, opened 2016-08-17
Performance Issue


timbuchwaldt:
As the readme doesn't really state it: django-rest-knox has a major performance issue.

https://github.com/James1345/django-rest-knox/blob/master/knox/auth.py#L50

This line leads to _major_ performance hits once one reaches significant user numbers, as each API request leads to all tokens being sent do the Django instances. Once reaching a few tens of thousands of users and more than a handful of requests per second, the application is spending more time receiving/parsing SQL and guessing tokens than doing some actual work.

One way to change this would be the following: Split the token in 2 parts:
- Part one is an identifier for a database column,  e.g.  a random string
- Part two is the very token knox currently uses

This would lead to only querying for a single row (the one matching part on), then comparing one hash.

Funfact: While loadtesting I watched my servers handle > 100mbit/s of SQL due to this bug.

jerzyk:
@timbuchwaldt agreed, this is an real issue, especially when your user base is growing and those users are opening multiple sessions

I was thinking how to limit a subset of the data that is fetched from the database e.g. by filtering only active tokens or by extending model and include e.g. ip address and filter in the first place all tokens with specific ip

this is similar to your approach, but then we are loosing one functionality - self cleaning of the token table

finally what we have implemented here is pretty simple: we are using cache to resolve tokens:
- when user is accessing site first time, his token is not in cache, so there is call to the database, when successful - cache is being set
- subsequent calls are getting data from the cache itself, very fast as this is direct key access

this is very simple solution and save us a lot of hassle

one more place that can improve performance is setting cache after successful login (at token creation), to not delay cache setting till first token request,but we did not implemented it.

there was one more idea - to hash a keys in the cache itself - e.g. to get md5 or sha-xxx of the token and use it as a key in the cache - but in our situation it would not bring any additional security level and will put some more load, delaying whole process.

timbuchwaldt:
Limiting to just a subset of data is basically what I proposed (and @raphaa implemented in a fork), by just adding a second column to contain part of the token. Limiting to IPs has obvious problems (e.g. taking your device to another place).

Self cleaning seems like a nice property, but this could easily be implemented as a task run by cron, celery beat or similar systems. Heck, even running `DELETE FROM tokens WHERE created_at  < DATEADD(day, -30, GETDATE()` at a percentage of requests would do the trick. But passing arround *all* data on *every* request is just slow.

[...]

PR #29, opened 2016-12-01
Performance improvements part II

raphaa:
This is the second part, with a newer version. Prevents a break with existing installations. Unfortunately, this contains the commit of the other fork.


belugame:
@raphaa do your tests work on your end?

ryanschneider:
I was just checking out this PR, as we're starting to hit the performance issues during peak daily traffic.  

One suggestion: what if instead of deleting all the tokens the logic for `authenticate_credentials(self, token)` became:

```python
def authenticate_credentials(self, token):
        # first check tokens with matching token_key
        for auth_token in AuthToken.objects.filter(
                token_key=token[:CONSTANTS.TOKEN_KEY_LENGTH]):
            if auth_token.expires is not None:
                if auth_token.expires < timezone.now():
                    auth_token.delete()
                    continue
            digest = hash_token(token, auth_token.salt)
            if digest == auth_token.digest:
                return self.validate_user(auth_token)

        # then check tokens that haven't been migrated
        for auth_token in AuthToken.objects.filter(token_key=''):
            if auth_token.expires is not None:
                if auth_token.expires < timezone.now():
                    auth_token.delete()
                    continue
            digest = hash_token(token, auth_token.salt)
            if digest == auth_token.digest:
                return self.validate_user(auth_token)

        # Authentication with this token has failed
        raise exceptions.AuthenticationFailed(_('Invalid token.'))
```

(I keep the code in the loop duplicated here for clarity, but would probably refactor the block into a function).

Basically, as tokens expire and are deleted, tokens with a blank `token_key` should become increasingly rare, and eventually non-existent, so the 2nd loop becomes a single fast query.  But, IMO, this would no longer be a "breaking" change and could be considered 2.3.0 instead of 3.0 since no tokens would be harmed during the upgrade process.

Thoughts?

James1345:
@ryanschneider @belugame @raphaa 
I'm sure I've said it before somewhere but I'll say it again in reference to your comment @ryanschneider ,
I do not consider the deletion of existing tokens to be a breaking change. Harming tokens during upgrade is not breaking *provided that* the API is not changed. This is meant to be used with expiring tokens in a RESTful scenario, so all clients (Single Page Applications, Apps, Other services drawing data via Http) *should* be capable of re-authenticating using the LoginView (either by using stored credentials or asking a user to authenticate again)
I.E. Tokens are the REST analog of Sessions - not something expected to be persistent for any great length of time. (In my systems tokens have an expiry of 10 hours - forcing users to re-enter their password. Prematurely deleting them all will cause no issues as the user will simpluy re-auth)

ryanschneider:
I noticed in this PR the version was 3.0, is that still going to be the case then?  As far as I can tell there aren't any breaking API changes, so assumed it was 3.0 because of the tokens being deleted.

On a related note, is there anything I can do to help get this PR accepted and a new release cut?  Would figuring out why the Travis tests are failing help?  I looked briefly and saw it has to do with the migration, but can spend more time on it tomorrow if it'd help move along the next release.

ryanschneider:
@James1345 @raphaa I spent a bit of time looking into the test failures, I currently see 2 failing tests:

- The first was just an issue with the structure of the `test_update_token_key` test, I fixed the implementation of that test in this commit: https://github.com/ryanschneider/django-rest-knox/commit/968c77af7861d40e8d8427d63a1ede5eede7aa51

- The second failure is in `test_expired_tokens_deleted` and actually exposes a bigger issue: Currently logout does **not** delete all tokens for a user, it seems to only delete at most a single token.  I've updated the `test_logout_deletes_keys` test to create multiple tokens to make this more explicit in this commit: https://github.com/ryanschneider/django-rest-knox/commit/9a6ba16b448b6379c9ead0deec171539c9dc167a

I'll take a stab at determining why all tokens are not deleted and report back with my findings and/or a fix depending on the complexity of the issue.

ryanschneider:
Oh wait, I was wrong: I was confusing `knox_logout` and `knox_logoutall`.  I'll update my change to `test_logout_deletes_keys` to ensure only the current token is deleted, and then focus on why expired tokens aren't being deleted.

But that raises a question: do we want to keep the behavior that expired tokens are deleted, or are they acceptable to keep around now that the entire table isn't being iterated over?

ryanschneider:
Ok, I added a bit of code to delete a user's other expired tokens when they attempt to authenticate in this commit: https://github.com/ryanschneider/django-rest-knox/commit/67bd3793b2786879f7b1e3795ff4e714c8e34dcb

I also added two more tests just for my own sanity:

- Verify attempting to authenticate with an expired token fails
- Verify that logout all only deletes the tokens for the current user

Both obviously pass, and I updated the `test_logout_deletes_keys` test to verify that only the current token is deleted.

With these changes my tests pass locally.  Would you like me to create a PR?

belugame:
@ryanschneider  Contributions are always highly welcome, happy to see a pull request from your side!

raphaa:
Sorry guys for my late response. Was far away from this issue here.
So… Why is this 3.0? Because with these changes, all users who didn't consumed the API with 2.3.0 (see #28) will have invalid tokens. You also could use 2.4.0 but my intention was to protect developers who use "django-rest-knox<3.0" in their requirements.

Thanks @ryanschneider for fixing the tests. In my opinion the code shouldn't delete any tokens. If a user created e.g. 10 tokens once but never consumed the API afterwards, the tokens never will be deleted. So like I've said in #21, I highly recommend a dedicated cron job/celery task to clean the table. This is also what Django does with `manage.py clearsessions`.

ryanschneider:
Ok, I've submitted my PR #39, with these high-level changes:

- Added logic to delete expired tokens for a user.  I'm not against the 'cron' approach but wanted the current tests to pass since they seem to be the best representation of a contract for the code.
- Added a couple more tests to make sure things behave correctly (logging out only deletes 1 token, and that attempting to use an expired token fails).

@raphaa can you point me towards any more discussion on what you mean by "invalid tokens".  Is there an upgrade scenario where either the migration fails or users cannot log in, or would there just be some garbage tokens in the DB?

Speaking of upgrades, I also want to verify that a user can perform the DB migration from 2.x -> 3.0 and still use the 2.x code without issue, as I for one don't want to force downtime for the upgrade.

James1345:
Having read through the discussions, I agree that for now we leave the code as-is re deleting tokens. My favorite option that's come out of all of this is to, in the future, remove this function, and have it in a Management command that the users can schedule if they want - reason being I would like an in-project way to clean out the tokens, but I also like to give the users their own choice of scheduling tool.

PR #39, opened 2017-01-12
Update to @raphaa's performance improvements.

ryanschneider:
The changes from #29 include:

- Merged in latest master
- Updated the test `test_expired_tokens_deleted` to more accurately reflect how tokens are generated
- Add code to `TokenAuthentication.authenticate_credentials` to delete expired tokens for a user so that the test `test_expired_tokens_deleted` passes.
- Added code to `test_logout_deletes_keys` to ensure only the logged out key is deleted.
- Added `test_logout_all_deletes_only_targets_keys` and `test_expired_tokens_login_fails ` tests.


ryanschneider:
I'm testing upgrading from 2.2 to 3.0 today.  I'll hopefully have a PR today with some minor changes that ease the upgrade process.

ryanschneider:
Actually, n/m, there won't be a PR.  I was thinking that adding a default value to `token_key` would ease the migration, but I forgot that Django doesn't actually set default values at the DB level, so the migration doesn't actually help (the user needs to manually add a default value of '' if they want to do the DB upgrade separate from the code upgrade).

Anyways, good news is that my testing today has been very promising, so I'm planning on pushing this branch live for us some time this week.  I'll definitely report back on any issues and how it goes.

ryanschneider:
Ok, we went live with this version this morning, and can see the performance pay off already.

The only hitch we had is we have an older client that fails to handle the 4xx results correctly and re-auth, but that's not an issue with Knox (the client re-auths fine if it is restarted so we were willing to accept that). 

IMO, the only barrier to releasing 3.0 is updating the docs to document that migration is required and that clients will need to re-auth as all tokens are deleted.

belugame:
sorry for the delay, I will prepare the release notes and see how we can best make users make aware of the required migration

ryanschneider:
Glad to hear it!  We've been running my branch in production now for 3 weeks and average latency for auth requests has gone from around 1s (typically 400ms but with spikes of up to 2-3s under load) to around 40ms with few if any spikes.

PR #65, opened 2017-08-13
Fix SyntaxError in settings example in documentation

fmichea:
There seems to be a small syntax error with the settings example where the documentation recommends assigning to a literal string. This commit just removes the quotes to make it a variable name.


belugame:
thank you :+1:  I will merge it to the develop branch first

PR #69, opened 2017-09-14
Use os.urandom instead of OpenSSL.rand.bytes

philngo:
Follows suggestion in pyOpenSSL changelog https://github.com/pyca/pyopenssl/blob/1eac0e8f9b3829c5401151fabb3f78453ad772a4/CHANGELOG.rst#backward-incompatible-changes-1

### DIFF
--- a/knox/crypto.py
+++ b/knox/crypto.py
@@ -2,7 +2,7 @@ import binascii
 
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes
-from OpenSSL.rand import bytes as generate_bytes
+from os import urandom as generate_bytes
 
 from knox.settings import knox_settings, CONSTANTS
 


AndrewPix:
Maybe we should use Python3.6 secrets module to generate cryptographically strong token?

> Starting with Python 3.6 the standard library includes the secrets module, which can be used for generating cryptographically secure random numbers, with specific helpers for text-based formats.

fahdk:
@AndrewPix https://github.com/python/cpython/blob/master/Lib/secrets.py
secrets.token_bytes is just a wrapper around os.urandom

007:
Ref https://github.com/pyca/pyopenssl/pull/675

belugame:
@James1345 has deliberately avoided using urandom, so I would like to hear his opinion on how to step forward here: 

From the Readme:

> DRF Tokens are generated with os.urandom, which is not cryptographically secure.

James1345:
I did originally avoid urandom in favour of OpenSSL's random, as that was the "best" cross platform random at the time. Now it is agreed that urandom is the correct way (which is why the OpenSSL method is being dropped) so we'll use that

James1345:
Aaaaand I didn't realize that was done on master so have just merged that in completely the wrong way.

Issue #77, opened 2017-12-14
Tests fail on Django 2.0


belugame:
see build output of Django 2.0 here https://travis-ci.org/James1345/django-rest-knox/builds/316640570?utm_source=github_status&utm_medium=notification

belugame:
The token login request in the tests already returns a 401 response. Might be a str/unicode issue on python3

belugame:
closed with release 3.1.0

PR #78, opened 2017-12-15
Django 2.0 update

vkylamba:


belugame:
Hey @vkylamba thank you very much for your contribution. I think you branched off an outdated master commit or so. We need to work with the latest development branch to use the new tox setup.

belugame:
I changed the target branch of your merge request to develop. We definitely need a solution that supports both python2.7/Django 1.11 and python3/Django 2.0

belugame:
closes #77

Issue #63, opened 2017-06-23
Need more information in setup knox docs


ishitap:
In the [docs (Installation > Setup)](https://james1345.github.io/django-rest-knox/installation/#setup-knox), it says to "Make knox's TokenAuthentication your default authentification class for django-rest-framework." 

[Only way later in the docs](https://james1345.github.io/django-rest-knox/auth/) does it mention that doing so will require you to override the login view. 

It took me forever to figure out how I was getting a 401 even though I followed all the setup instructions. I think the information about overriding should be included in the setup, or there should be different setup instructions altogether so people can run a test quickly and make sure everything works post-install.

belugame:
Yes, that should be included at least. Do you want to propose additional instructions that would have helped you?

PR #81, opened 2018-01-15
updates docs to better explain the use of TokenAuthentication

hakur0:
As explained by @ishitap on #63, the docs needed a little more information on the matter. This should better explain the situation to newcomers.


belugame:
Thank you very much for your contribution, @hakur0 Lucas!

PR #84, opened 2018-01-25
Safer digests comparison

### LINKED ISSUE #83: don't compare digests directly
https://github.com/James1345/django-rest-knox/blob/e2c59581eac21381f9e789494744f20e4e187e7b/knox/auth.py#L71

it's bad to use `==` directly

https://docs.python.org/2/library/hmac.html#hmac.compare_digest

fengsi:
closes #83


belugame:
Thank you again for your prompt contribution! :+1:

PR #90, opened 2018-02-25
Add fix and test for binascii.Error Odd-Length

### LINKED ISSUE #46: Odd token length trigger a "Odd-length string" server error
Version: django-rest-knox==2.3.0

**Produces an error 401:**
- Authorization: Token 12
- Authorization: 123

**Produces an "Odd-length string" error 500:**
- Authorization: Token 123
- Authorization: Token abcde
- ... every other token with an odd length

chdsbd:
This fixes an edge case that only occurs when valid token is extended.

If I take a token a valid token (even length) and add a character to it, this will make an invalid, odd-length token. This token will pass the [first check][0], because the token is sliced to the correct length in the filter statement, but then an error occurs in [knox.crypto][1].

Fixes #46 

[0]: https://github.com/James1345/django-rest-knox/blob/45be9d1fc0c7963954803a3b01425f6b540f8673/knox/auth.py#L64

[1]: https://github.com/James1345/django-rest-knox/blob/45be9d1fc0c7963954803a3b01425f6b540f8673/knox/crypto.py#L29-L30

### DIFF
--- a/knox/auth.py
+++ b/knox/auth.py
@@ -4,6 +4,8 @@ except ImportError:
     def compare_digest(a, b):
         return a == b
 
+import binascii
+
 from django.conf import settings
 from django.utils.translation import ugettext_lazy as _
 from django.utils import timezone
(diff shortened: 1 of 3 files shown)


belugame:
Thank you for the good work! I will prepare a bugfix release

PR #95, opened 2018-04-06
Refactor views to subclass APIView instead of GenericAPIView

### LINKED ISSUE #94: Views do not have a serializer_class
I am attempting to use django-rest-swagger for API documentation with DRF 3.8.1. Loading fails with the following exception:

```
'LoginView' should either include a `serializer_class` attribute, or override the `get_serializer_class()` method.
```

In the DRF documentation, it specifies that:
> `serializer_class` - [...] Typically, you must either set this attribute, or override the get_serializer_class() method.

http://www.django-rest-framework.org/api-guide/generic-views/#attributes

If I'm understanding the documentation correctly, it appears that DRF `GenericAPIView` must have a `serializer_class`, which `LoginView`, `LogoutView`, and `LogoutAllView` do not, so I believe this is a bug in django-rest-knox and not in django-rest-swagger. 

I'm going to override the classes in my project to add a `serializer_class`, but I think it might be a fairly involved set of changes. If you're interested in a pull request, please let me know and I can work on creating one once I make the changes.

jonpavelich:
This pull request refactors the three Knox views to subclass `APIView` instead of `GenericAPIView`, which resolves #94. 

The only `GenericAPIView` functionality currently being used is `get_serializer_context()` in `LoginView`, which is a very simple function that creates a context dictionary with no other `GenericAPIView` dependencies. Its functionality has been replicated by creating the context dictionary directly in `LoginView`.

### DIFF
--- a/knox/views.py
+++ b/knox/views.py
@@ -3,14 +3,14 @@ from rest_framework import status
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 from rest_framework.settings import api_settings
-from rest_framework.generics import GenericAPIView
+from rest_framework.views import APIView
 
 from knox.auth import TokenAuthentication
 from knox.models import AuthToken
 from knox.settings import knox_settings
 
 
-class LoginView(GenericAPIView):
+class LoginView(APIView):
     authentication_classes = api_settings.DEFAULT_AUTHENTICATION_CLASSES
     permission_classes = (IsAuthenticated,)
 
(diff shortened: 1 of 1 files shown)


clintonb:
@James1345 @belugame would you be open to undoing this change, and going back to `GenericAPIView`? Doing so would enable me to generate proper OpenAPI schema without having to override the class.

belugame:
@clintonb Sorry, I m no longer actively maintaining here

danihodovic:
@clintonb Could you open a PR for this? I'm having the same problem.

clintonb:
@danihodovic more than two years have elapsed. 

```

**Comment by clintonb** [comment]
@danihodovic more than two years have elapsed. I’m not opening a pull request, but this is the code I used to workaround this issue. 

```python
# NOTE: This customization exists so that we generate the appropriate schema for OpenAPI
class KnoxLoginView(LoginView, GenericAPIView):
    serializer_class = serializers.LoginSerializer
```

adinhodovic:
Also works:

```py
SPECTACULAR_SETTINGS = {
    "VERSION": "1.0.0",
    "SERVE_PERMISSIONS": ["rest_framework.permissions.IsAdminUser"],
    "APPEND_COMPONENTS": {
        "securitySchemes": {
            "ApiKeyAuth": {"type": "apiKey", "in": "header", "name": "Authorization"}
        }
    },
    "SECURITY": [
        {
            "ApiKeyAuth": [],
        }
    ],
```

Issue #98, opened 2018-04-17
IsAuthenticated Permission on LoginView


cwkang1998:
Forgive me if I misunderstand the implementation for the LoginView for knox, but shouldn't the LoginView have a permission of 'Allow Any' instead of 'IsAuthenticated' ?

belugame:
No worries, that behavior indeed looks illogical on first read as compared to Django's.
Maybe we should reconsider the naming...

Think of LoginView as CheckCredentialsAndReturnATokenView. You need to send username and password via e.g. Basic-Auth headers to that view and it will return a token to you if they match an existing user. 

Now if you had AllowAny, it would give everyone a token without having to provide any sort of credentials, which makes authentication redundant.

cwkang1998:
@belugame 
Thank you for your reply, a few further questions here and some confirmation:

> Think of LoginView as CheckCredentialsAndReturnATokenView. You need to send username and password via e.g. Basic-Auth headers to that view and it will return a token to you if they match an existing user.

Since I am using KnoxTokenAuth as the default authentication, am I supposed to change it to BasicAuth from django-rest for authentication specifically for this endpoint and explicitly state that other endpoints will be using KnoxTokenAuth? If so, it is not stated in the documentation (but it is stated in Django Rest Authentication Page) so its a little bit confusing.

I am currently implementing my own LoginView by extending the KnoxLoginView to login using credentials directly (providing user and password and receiving a token without first going through a BasicAuthentication) while still setting KnoxTokenAuth as my default authentication.

cwkang1998:
For your reference, here is my custom implementation:

`from django.contrib.auth import login

from rest_framework import permissions
from rest_framework.authtoken.serializers import AuthTokenSerializer
from knox.views import LoginView as KnoxLoginView

class LoginView(KnoxLoginView):
    permission_classes = (permissions.AllowAny,)

    def post(self, request, format=None):
        serializer = AuthTokenSerializer(data=request.data)
        serializer.is_valid(raise_exception=True)
        user = serializer.validated_data['user']
        login(request, user)
        return super(LoginView, self).post(request, format=None)`

belugame:
I'm not sure what you are trying to achieve... what do you send to your LoginView? A token? I would expect username/password.

> providing user and password and receiving a token without first going through a BasicAuthentication

Providing username/password and getting a token is what you can do with BasicAuthentication. Maybe you were thinking that is an extra step.
[...]

PR #100, opened 2018-07-09
Added a new section in auth docs

cwkang1998:
this aims to provide a example implementation for devs that aim to use Knox Token Authentication as their only authentication method.


cwkang1998:
This is a follow up pull request on issue #98

belugame:
Thank you very much @cwkang1998, it is online now

Issue #37, opened 2016-12-26
Recommended method of refreshing tokens


cdriehuys:
I would like to implement a method of refreshing tokens from a single page app. The refresh itself seems pretty simple, but I was wondering what the best way of accessing a token's expiration time is. I was thinking about overriding the login view to return the expiration time as well, but the manager method to create new tokens only returns the token's key. Any input would be much appreciated.

belugame:
I'm not sure if I understand your problem correctly. As you suggested I would go with overriding the `post` method of the `LoginView` and simply add `'expiry': token.expiry` to the returned dictionary at https://github.com/James1345/django-rest-knox/blob/develop/knox/views.py#L23.

cdriehuys:
Thanks for the response. However, the custom manger used for the AuthToken model only returns the token string as seen [here](https://github.com/James1345/django-rest-knox/blob/master/knox/models.py#L23-L24).

belugame:
Ah, should have read more carefully. I can't see a reason why `create` shouldn't return the AuthToken instance and it should not need much changes in the codebase either. So if you'd like to make a pull request...

Otherwise I would suggest just looking up the AuthToken record by the token string, the performance impact should be minimal.

cdriehuys:
I messed around with the codebase a little, and it appears that a change like that is not so simple. For starters, the `token` generated [here](https://github.com/James1345/django-rest-knox/blob/master/knox/models.py#L13) is not equivalent to the `token_key` property on the model generated later on in that method [here](https://github.com/James1345/django-rest-knox/blob/master/knox/models.py#L20). Since the token that is currently returned from the `create` method is necessary for authentication, it is crucial that the `LoginView` is able to return that token.

If you have any more ideas on how to get the same token value generated [earlier in the function](https://github.com/James1345/django-rest-knox/blob/master/knox/models.py#L13) from an `AuthToken` instance, I would be happy to make a pull request.

Thanks for your time.

dhbradshaw:
What if create returned an authtoken object with a token property?  The token property would be ephemeral, but could be accessed for the api.

This would bring the create method return value back to a more standard form while still allowing access to the undigested token.

belugame:
close due to inactivity

muelli:
We are currently using

```
class KnoxTokenAuthentication(TokenAuthentication):
    def authenticate(self, request):
        try:
            result = super().authenticate(request)
            if result:
                user, auth_token = result
[...]

PR #104, opened 2018-07-24
Add AUTO_REFRESH setting

rphlo:
Implement enhancement https://github.com/James1345/django-rest-knox/issues/37

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -14,6 +14,7 @@ REST_KNOX = {
   'AUTH_TOKEN_CHARACTER_LENGTH': 64,
   'TOKEN_TTL': timedelta(hours=10),
   'USER_SERIALIZER': 'knox.serializers.UserSerializer',
+  'AUTO_REFRESH': FALSE,
 }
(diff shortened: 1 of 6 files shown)


belugame:
@rphlo thanks a lot for this feature. I m happy to merge it in. 2 things. 
a) we need unit tests that cover it e.g. assert that the token expiry timestamp is (not) extended when expected. 
b) I would make the 1 second threshold bigger to avoid hitting the database more than it is useful for this usecase. 5 or even 10 seconds seems reasonable for me

rphlo:
@belugame what about using freezegun in tests? This would add a dependency but would bring many benefit in my point of view for the tests...

belugame:
@rphlo sure, I m familiar with freezegun. Feel free to add it to tox.ini

belugame:
@rphlo Thank you for the work again, I extended your unit tests and tweaked things elsewhere unrelated to your changes. Glad to have your review my changes. https://github.com/James1345/django-rest-knox/pull/105

Issue #108, opened 2018-08-09
[Discussion/Suggestion] Ability to limit the amount of active tokens per user


sphrak:
Hello,
Im testing out this library and so far I am really liking it, but I wonder if there is a way to throttle or limit the amount of keys a user can have?

I am willing to take a stab at this if it does not exist, but im curious to hear what everyone else thinks about it?

EDIT: I think that probably a limit is the way to go rather than throttling the requests since that can be handled by the throttling mechanism already built into DRF.

Kindly,
Sphrak

rphlo:
Not sure what would be the use case for this but I would go by overriding the LoginView

```
class TokenLimitingLoginView(LoginView):
    max_token_count_per_user = 10
    def post(self, request, format=None):
        user_token_count = request.user.auth_token_set.all().count()
        if user_token_count >= self.max_token_count_per_user:
            return Response(
                {'error': 'User has reached max number of token per user'},
                status=status.HTTP_401_UNAUTHORIZED
            )
        super().post(request, format)
```

belugame:
@sphrak thank your for your feature suggestion.

FYI We do clean up expired tokens here https://github.com/James1345/django-rest-knox/blob/5ace943e46c799da5c9eeb2e65336d88e63ca008/knox/auth.py#L103

You are right that it does not stop users from creating a large number of tokens. For most usecases a small number of tokens (if not 1 per user) would suffice. So I am not against implementing such limit which one can override per setting.

Please remember to also cover it with unit testing when you want to have a go at it.

sphrak:
I close this issue since this has been accepted and now only merge into master is left. Discussion continues in PR if needed.

PR #109, opened 2018-08-09
add: optional token limit per user object

sphrak:
@belugame This is what I got thus far. I have verified the functionality manually -- it works but I am having trouble with the test not respecting the overridden `TOKEN_LIMIT`, thus the test fails -- I figured the default settings value should be `None` since this option does come with a performance cost and is not needed by everyone.

I mean I could just set `knox_settings.TOKEN_LIMIT = 10` inside the testcase but.. is that the correct way?

Related: https://github.com/James1345/django-rest-knox/issues/108

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -14,6 +14,7 @@ REST_KNOX = {
   'AUTH_TOKEN_CHARACTER_LENGTH': 64,
   'TOKEN_TTL': timedelta(hours=10),
   'USER_SERIALIZER': 'knox.serializers.UserSerializer',
+  'TOKEN_LIMIT_PER_USER': None,
   'AUTO_REFRESH': FALSE,
 }
 #...snip...
@@ -53,6 +54,10 @@ Setting the TOKEN_TTL to `None` will create tokens that never expire.
 
 Warning: setting a 0 or negative timedelta will create tokens that instantly expire,
 the system will not prevent you setting this.
+
+## TOKEN_LIMIT_PER_USER
+This allows you to control how many tokens can be issued per user.
+By default this option is disabled and set to `None` -- thus no limit.
 
 ## USER_SERIALIZER
 This is the reference to the class used to serialize the `User` objects when
--- a/knox/views.py
+++ b/knox/views.py
@@ -1,4 +1,7 @@
 from django.contrib.auth.signals import user_logged_in, user_logged_out
+from django.utils import timezone
+from django.conf import settings
+
 from rest_framework import status
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
@@ -15,6 +18,12 @@ class LoginView(APIView):
     permission_classes = (IsAuthenticated,)
 
     def post(self, request, format=None):
+        if knox_settings.TOKEN_LIMIT_PER_USER is not None:
+            if request.user.auth_token_set.filter(expires__gt=timezone.now()).count() >= knox_settings.TOKEN_LIMIT_PER_USER:
+                return Response(
+                    {"error": "Maximum amount of tokens allowed per user exceeded."},
+                    status=status.HTTP_403_FORBIDDEN
+                )
         token = AuthToken.objects.create(request.user)
         user_logged_in.send(sender=request.user.__class__, request=request, user=request.user)
         UserSerializer = knox_settings.USER_SERIALIZER
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -4,7 +4,7 @@ from datetime import datetime, timedelta
 from django.utils.six.moves import reload_module
 from django.contrib.auth import get_user_model
 from django.test import override_settings
-from knox import auth
+from knox import auth, views
 
 try:
     # For django >= 2.0
@@ -38,6 +38,11 @@ def get_basic_auth_header(username, password):
 auto_refresh_knox = knox_settings.defaults.copy()
 auto_refresh_knox["AUTO_REFRESH"] = True
 
+token_user_limit_knox = knox_settings.defaults.copy()
+token_user_limit_knox["TOKEN_LIMIT_PER_USER"] = 10
+
+user_serializer_knox = knox_settings.defaults.copy()
+user_serializer_knox["USER_SERIALIZER"] = UserSerializer
 
 class AuthTestCase(TestCase):
 
@@ -73,15 +78,18 @@ class AuthTestCase(TestCase):
         self.assertNotIn(username_field, response.data)
 
     def test_login_returns_serialized_token_and_username_field(self):
-        self.assertEqual(AuthToken.objects.count(), 0)
-        url = reverse('knox_login')
-        self.client.credentials(
-            HTTP_AUTHORIZATION=get_basic_auth_header(self.username, self.password)
-        )
-        knox_settings.USER_SERIALIZER = UserSerializer
-        response = self.client.post(url, {}, format='json')
+
+        with override_settings(REST_KNOX=user_serializer_knox):
+            reload_module(views)
+            self.assertEqual(AuthToken.objects.count(), 0)
+            url = reverse('knox_login')
+            self.client.credentials(
+                HTTP_AUTHORIZATION=get_basic_auth_header(self.username, self.password)
+            )
+            response = self.client.post(url, {}, format='json')
+            self.assertEqual(user_serializer_knox["USER_SERIALIZER"], UserSerializer)
+        reload_module(views)
         self.assertEqual(response.status_code, 200)
-        self.assertNotEqual(knox_settings.USER_SERIALIZER, None)
         self.assertIn('token', response.data)
         username_field = self.user.USERNAME_FIELD
         self.assertIn('user', response.data)
@@ -254,3 +262,37 @@ class AuthTestCase(TestCase):
 
         self.assertTrue(self.signal_was_called)
 
+    def test_exceed_token_amount_per_user(self):
+
+        with override_settings(REST_KNOX=token_user_limit_knox):
+            reload_module(views)
+            for _ in range(10):
+                token = AuthToken.objects.create(user=self.user)
+            url = reverse('knox_login')
+            self.client.credentials(
+                HTTP_AUTHORIZATION=get_basic_auth_header(self.username, self.password)
+            )
+            response = self.client.post(url, {}, format='json')
+        reload_module(views)
+        self.assertEqual(response.status_code, 403)
+        self.assertEqual(response.data, {"error": "Maximum amount of tokens allowed per user exceeded."})
+
+    def test_does_not_exceed_on_expired_keys(self):
+
+        with override_settings(REST_KNOX=token_user_limit_knox):
+            reload_module(views)
+            for _ in range(9):
+                token = AuthToken.objects.create(user=self.user)
+            AuthToken.objects.create(user=self.user, expires=timedelta(seconds=0))
+            # now 10 keys, but only 9 valid so request should succeed.
+            url = reverse('knox_login')
+            self.client.credentials(
+                HTTP_AUTHORIZATION=get_basic_auth_header(self.username, self.password)
+            )
+            response = self.client.post(url, {}, format='json')
+            failed_response = self.client.post(url, {}, format='json')
+        reload_module(views)
+        self.assertEqual(response.status_code, 200)
+        self.assertIn('token', response.data)
+        self.assertEqual(failed_response.status_code, 403)
+        self.assertEqual(failed_response.data, {"error": "Maximum amount of tokens allowed per user exceeded."})
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1,4 +1,7 @@
-#Changelog
+# Changelog
+
+## 3.4.0
+- adds optional token limit
 
 ## 3.3.1
 - Ensure compatibility with Django 2.1 up to Python 3.7
--- a/docs/index.md
+++ b/docs/index.md
@@ -13,7 +13,8 @@ in to DRF. However, it overcomes some problems present in the default implementa
 
     Knox provides one token per call to the login view - allowing
     each client to have its own token which is deleted on the server side when the client
-    logs out.
+    logs out. Knox also provides an optional setting to limit the amount of tokens generated
+    per user.
 
     Knox also provides an option for a logged in client to remove *all* tokens
     that the server has - forcing all clients to re-authenticate.
--- a/knox/settings.py
+++ b/knox/settings.py
@@ -10,6 +10,7 @@ DEFAULTS = {
     'AUTH_TOKEN_CHARACTER_LENGTH': 64,
     'TOKEN_TTL': timedelta(hours=10),
     'USER_SERIALIZER': None,
+    'TOKEN_LIMIT_PER_USER': None,
     'AUTO_REFRESH': False,
 }
 


sphrak:
Oh doh you are absolutly right. My bad, totally missed that. I will fix it to only count unexpired keys.

knox/views.py
```diff
@@ -16,8 +18,11 @@ class LoginView(APIView):
 
     def post(self, request, format=None):
         if knox_settings.TOKEN_LIMIT is not None:
-            if (request.user.auth_token_set.count() >= knox_settings.TOKEN_LIMIT):
-                return Response({"error": "Maximum amount of tokens allowed per user exceeded."}, status=status.HTTP_401_UNAUTHORIZED)
+            if (request.user.auth_token_set.filter(expires__gt=timezone.now()).count() >= knox_settings.TOKEN_LIMIT):
```
rphlo: unnecessary parenthesis around if statement, also I would write tests for making sure expired token are properly filterered out.

tests/tests.py
```diff
@@ -205,10 +205,11 @@ def test_token_expiry_is_not_extended_within_MIN_REFRESH_INTERVAL(self):
         self.assertEqual(response.status_code, 200)
         self.assertEqual(original_expiry, AuthToken.objects.get().expires)
 
-    @override_settings(REST_KNOX=no_auto_refresh_knox)
+    #@override_settings(REST_KNOX=no_auto_refresh_knox)
```
rphlo: if you want to use `override_settings` you should not use knox_settings in the login view but `settings.REST_KNOX['TOKEN_LIMIT']` as it is done for `AUTO_REFRESH` setting in auth.py.
Then I would not use `no_auto_refresh_knox` but a new `token_limit_knox` settings.

sphrak: As stated previously I didnt get the override setting to work -- because if I try to evaluate `settings.REST_KNOX['TOKEN_LIMIT']` in a print statement thats fine, but if I try to do it in an `if` statement it becomes a `KeyError`. I can push it if you want to take a look.

sphrak: Ill just paste the code here:

```python
        if settings.REST_KNOX["TOKEN_LIMIT"] is not None:
            if request.user.auth_token_set.filter(expires__gt=timezone.now()).count() >= settings.REST_KNOX["TOKEN_LIMIT"]:
                return Response(
                    {"error": "Maximum amount of tokens allowed per user exceeded."},
                    status=status.HTTP_401_UNAUTHORIZED
                )
```

The first if statement will fail with a `KeyError`, if I instead check `knox_settings['TOKEN_LIMIT']` that will work, but then its always `None` since its a default value.

rphlo: I am away from computer for today, Might look at it after that.

docs/settings.md
```diff
@@ -54,6 +55,13 @@ Setting the TOKEN_TTL to `None` will create tokens that never expire.
 Warning: setting a 0 or negative timedelta will create tokens that instantly expire,
 the system will not prevent you setting this.
 
+## TOKEN_LIMIT
```
belugame: I'd prefer a name that does not imply there is a general token limit but emphasizes a per-user token limit.

knox/views.py
```diff
@@ -15,6 +18,12 @@ class LoginView(APIView):
     permission_classes = (IsAuthenticated,)
 
     def post(self, request, format=None):
+        if settings.REST_KNOX['TOKEN_LIMIT'] is not None:
```
belugame: please have a look at my changes here https://github.com/James1345/django-rest-knox/commit/c747f40a43fb1bd2c713183b1a49afbbcfdf60d0
I think it is the cleaner way of accessing settings

knox_project/settings.py
```diff
@@ -59,5 +59,6 @@
 TEST_RUNNER = 'django_nose.NoseTestSuiteRunner'
 
 REST_KNOX = {
-    'AUTO_REFRESH': True
+    'AUTO_REFRESH': True,
+    'TOKEN_LIMIT': None
```
belugame: Again I had set a bad example for this. But lets avoid having settings here as #111 has made me aware it is cleaner without.

docs/settings.md
sphrak: Like `TOKEN_LIMIT_PER_USER`? I will make sure to fix this once this has been decided https://github.com/James1345/django-rest-knox/pull/114.

sphrak:
I have fixed these issues in the latest commit -- however I am unable to run the tests locally since they depend on commits I do not have.. any chance you know what I could do @belugame? This PR is also blocked by merge conflict due to recent `3.2.1` merge into master.

rphlo:
@sphrak https://gist.github.com/ravibhure/a7e0918ff4937c9ea1c456698dcd58aa follow this guide to rebase your branch onto latest version

sphrak:
@rphlo Thank you, unfortunatly those instructions borked my local branch due to all the conflicts and has now diverged so I had to redo it. 
Weird, it really should just have been a matter of rebasing but its likely because I have more than 1 commit in this feature branch. Gonna try squash it then rebase to avoid the conflicts.

sphrak:
@belugame The merge conflict is fixed. However I cannot get the tests to pass with the new approach to test settings. Either im not checking the setting correctly in `views.py` or the `override_settings` does not work as I understood it.

_In a naive attempt I tried checking against `develop` branch, but to no avail so changed it back again._

Regards,
Sphrak

johnraz:
@sphrak : do you need some help to get this merged? I'd gladly help if that's the case.

tests/tests.py
```diff
@@ -209,3 +212,36 @@ def test_token_expiry_is_not_extended_within_MIN_REFRESH_INTERVAL(self):
 
         self.assertEqual(response.status_code, 200)
         self.assertEqual(original_expiry, AuthToken.objects.get().expires)
+
+    def test_exceed_token_amount_per_user(self):
+
+        with override_settings(REST_KNOX=token_user_limit_knox):
+            #reload_module(auth)
```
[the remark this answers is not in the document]
sphrak: You are right, that was indeed the problem. Thank you.

knox/views.py
```diff
@@ -15,9 +18,19 @@ class LoginView(APIView):
...
+        if knox_settings.TOKEN_LIMIT_PER_USER is not None:
+            if request.user.auth_token_set.filter(expires__gt=timezone.now()).count() >= knox_settings.TOKEN_LIMIT_PER_USER:
+                return Response(
+                    {"error": "Maximum amount of tokens allowed per user exceeded."},
+                    status=status.HTTP_401_UNAUTHORIZED
+                )
         token = AuthToken.objects.create(request.user)
         user_logged_in.send(sender=request.user.__class__, request=request, user=request.user)
         UserSerializer = knox_settings.USER_SERIALIZER
+        if UserSerializer is None:
```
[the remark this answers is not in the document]
sphrak: No, sorry it will be removed in next commit.

docs/settings.md
```diff
@@ -54,6 +55,13 @@ Setting the TOKEN_TTL to `None` will create tokens that never expire.
 Warning: setting a 0 or negative timedelta will create tokens that instantly expire,
 the system will not prevent you setting this.
 
+## TOKEN_LIMIT_PER_USER
+This allows you to control how many tokens that can be issued per user.
+By default this option is disabled and set to `None` -- thus no limit. 
+
+**WARNING:** This setting does not come without a performance cost -- one additional database
```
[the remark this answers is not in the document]
sphrak: I have removed the warning in next commit.

sphrak:
@johnraz I think I got it -- unless there is something else you want to add. I simply missed the notifications about the activity in this PR. I have fixed this PR now in latest commit.

Kindly

knox/views.py
```diff
@@ -15,9 +18,19 @@ class LoginView(APIView):
...
+                    {"error": "Maximum amount of tokens allowed per user exceeded."},
+                    status=status.HTTP_401_UNAUTHORIZED
+                )
         token = AuthToken.objects.create(request.user)
         user_logged_in.send(sender=request.user.__class__, request=request, user=request.user)
         UserSerializer = knox_settings.USER_SERIALIZER
+        #if UserSerializer is None:
+        #    return Response(
+        #        {'token': token}
+        #    )
```
rphlo: You could remove that commented code.

tests/tests.py
```diff
@@ -209,3 +212,36 @@ def test_token_expiry_is_not_extended_within_MIN_REFRESH_INTERVAL(self):
...
+            )
+            response = self.client.post(url, {}, format='json')
+            reload_module(views)
+            self.assertEqual(response.status_code, 401)
+            self.assertEqual(response.data, {"error": "Maximum amount of tokens allowed per user exceeded."})
+
+    def test_does_not_exceed_on_expired_keys(self):
+
+        with override_settings(REST_KNOX=token_user_limit_knox):
+            #reload_module(auth)
```
rphlo: Should have `reload_module(views)`  otherwise you are not testing your feature.
Would add a sign up request after the successful one to test that it indeed fail after 10 valid tokens.

sphrak: @rphlo my bad, I went to fast. What release do you want me to add this feature under in `changes.md`?

sphrak:
I suspect its failing because of this https://github.com/James1345/django-rest-knox/commit/7570615b6faaa741b1b5e5b0a76965047a2dd4c4 but that was added long time ago, I dont understand how its a problem now

This commit doesnt even touch that part

tests/tests.py
```diff
@@ -254,3 +257,36 @@ def handler(sender, username, **kwargs):
...
+        with override_settings(REST_KNOX=token_user_limit_knox):
+            reload_module(views)
+            for _ in range(10):
+                token = AuthToken.objects.create(user=self.user)
+            url = reverse('knox_login')
+            self.client.credentials(
+                HTTP_AUTHORIZATION=get_basic_auth_header(self.username, self.password)
+            )
+            response = self.client.post(url, {}, format='json')
+            reload_module(views)
```
rphlo: this should be indented outside `override_settings` block

tests/tests.py
```diff
@@ -254,3 +257,36 @@ def handler(sender, username, **kwargs):
...
+                token = AuthToken.objects.create(user=self.user)
+            AuthToken.objects.create(user=self.user, expires=timedelta(seconds=0))
+            # now 10 keys, but only 9 valid so request should succeed.
+            url = reverse('knox_login')
+            self.client.credentials(
+                HTTP_AUTHORIZATION=get_basic_auth_header(self.username, self.password)
+            )
+            response = self.client.post(url, {}, format='json')
+            failed_response = self.client.post(url, {}, format='json')
+        self.assertEqual(response.status_code, 200)
```
rphlo: `reload_module(views)` missing after the `override_settings` block

knox/views.py
```diff
@@ -15,6 +18,12 @@ class LoginView(APIView):
     permission_classes = (IsAuthenticated,)
 
     def post(self, request, format=None):
+        if knox_settings.TOKEN_LIMIT_PER_USER is not None:
+            if request.user.auth_token_set.filter(expires__gt=timezone.now()).count() >= knox_settings.TOKEN_LIMIT_PER_USER:
```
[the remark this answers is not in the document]
sphrak: I like the idea. But where would this check be you think? im not familiar with the startup procedure

tests/tests.py
sphrak: I have fixed this in latest commit.

tests/tests.py
sphrak: I have fixed this in latest commit.

johnraz:
@sphrak I just have some spare time and thought I could help if that was needed, nothing special I wanna add and if you sorted it then that's even better 🚀

johnraz:
> I suspect its failing because of this [7570615](https://github.com/James1345/django-rest-knox/commit/7570615b6faaa741b1b5e5b0a76965047a2dd4c4) but that was added long time ago, I dont understand how its a problem now
> 
> This commit doesnt even touch that part

@sphrak could it be that the reload of the `views` module is re-initializing the `knox_settings.USER_SERIALIZER` to `None`?

Maybe reloading views after https://github.com/James1345/django-rest-knox/blob/master/tests/tests.py#L81 would make it safer?

I'm just fishing here and throwing ideas though I'm not very familiar with that module reloading from `six`.

sphrak:
@belugame @rphlo @johnraz alright so I got it working but I am not sure if I like it because it modifies code that is sort of outside the scope of this commit and I have no clue why it is even working.

tests/tests.py
```diff
@@ -73,13 +76,17 @@ def test_login_returns_serialized_token(self):
...
     def test_login_returns_serialized_token_and_username_field(self):
-        self.assertEqual(AuthToken.objects.count(), 0)
-        url = reverse('knox_login')
-        self.client.credentials(
-            HTTP_AUTHORIZATION=get_basic_auth_header(self.username, self.password)
-        )
-        knox_settings.USER_SERIALIZER = UserSerializer
-        response = self.client.post(url, {}, format='json')
+
+        with override_settings(REST_KNOX=token_user_limit_knox):
```
rphlo: Would use a different settings set `user_serializer_knox`

tests/tests.py
```diff
@@ -73,13 +76,17 @@ def test_login_returns_serialized_token(self):
...
+            reload_module(views)
+            self.assertEqual(AuthToken.objects.count(), 0)
+            url = reverse('knox_login')
+            self.client.credentials(
+                HTTP_AUTHORIZATION=get_basic_auth_header(self.username, self.password)
+            )
+            knox_settings.USER_SERIALIZER = UserSerializer
```
rphlo: This line is not necesarry since you already overwrote the setting

sphrak: Thats where you're mistaken my friend. `knox_settings.USER_SERIALIZER = UserSerializer` is indeed still needed otherwise `USER_SERIALIZER` is `None` or I need to reload some other module perhaps

tests/tests.py
sphrak: Fixed in next commit.

tests/tests.py
```diff
@@ -73,15 +78,18 @@ def test_login_returns_serialized_token(self):
...
     def test_login_returns_serialized_token_and_username_field(self):
-        self.assertEqual(AuthToken.objects.count(), 0)
-        url = reverse('knox_login')
-        self.client.credentials(
-            HTTP_AUTHORIZATION=get_basic_auth_header(self.username, self.password)
-        )
-        knox_settings.USER_SERIALIZER = UserSerializer
-        response = self.client.post(url, {}, format='json')
+
+        with override_settings(REST_KNOX=user_serializer_knox):
```
johnraz: You could use the decorator as the override can be applied to the all test.

```
@override_settings(REST_KNOX=user_serializer_knox)
def test_login_returns_serialized_token_and_username_field(self):
...
```

This clarifies the fact that the settings can be overridden during the all lifetime of the test and also remove the indentation which makes the code slightly more readable.

This applies to the other usage below.
It's really just a cosmetic change, nothing blocking really.

tests/tests.py
```diff
@@ -73,15 +78,18 @@ def test_login_returns_serialized_token(self):
...
+            self.assertEqual(AuthToken.objects.count(), 0)
+            url = reverse('knox_login')
+            self.client.credentials(
+                HTTP_AUTHORIZATION=get_basic_auth_header(self.username, self.password)
+            )
+            response = self.client.post(url, {}, format='json')
+            self.assertEqual(user_serializer_knox["USER_SERIALIZER"], UserSerializer)
+        reload_module(views)
         self.assertEqual(response.status_code, 200)
-        self.assertNotEqual(knox_settings.USER_SERIALIZER, None)
```
johnraz: I think this should stay in place.

tests/tests.py
rphlo: I disagree on that point, as we need to reload the module (here `views`) after the override as well. Having the global override would make this more difficult.

tests/tests.py
rphlo: If this was to stay it should be moved within the override

docs/settings.md
```diff
@@ -54,6 +55,10 @@ Setting the TOKEN_TTL to `None` will create tokens that never expire.
 Warning: setting a 0 or negative timedelta will create tokens that instantly expire,
 the system will not prevent you setting this.
 
+## TOKEN_LIMIT_PER_USER
+This allows you to control how many tokens that can be issued per user.
```
belugame: @sphrak  the "that" should be deleted

knox/views.py
```diff
@@ -15,6 +18,12 @@ class LoginView(APIView):
...
+                return Response(
+                    {"error": "Maximum amount of tokens allowed per user exceeded."},
+                    status=status.HTTP_401_UNAUTHORIZED
```
belugame: To me it feels more like a 403, @sphrak 
What do you think? https://en.wikipedia.org/wiki/List_of_HTTP_status_codes

belugame:
Good work guys! Added 2 minor comments that could be addressed, but happy to merge this!

knox/views.py
sphrak: I agree @belugame I will fix your proposed changes in upcoming commit.

sphrak:
It has been fixed @belugame :-)

belugame:
for the 3.4.0 release we need to adapt the setup.py

Issue #107, opened 2018-07-31
Latest release 3.3.1 fails to deploy to pypi


belugame:
see job log: https://travis-ci.org/James1345/django-rest-knox/jobs/409780228

HTTPError: 400 Client Error: User 'xxx.xxx' does not have a verified primary email address. Please add a verified primary email before attempting to upload to PyPI. See https://pypi.org/help/#verified-email for more information.for more information. for url: https://upload.pypi.org/legacy/

Original author was informed by me.

belugame:
ping @James1345

belugame:
Feel free to drop him an email, you can find it here:
https://github.com/James1345/django-rest-knox/blob/master/setup.py#L28

rphlo:
Thing is he should have already received multiple email regarding this issue and did not respond to any of those (Pypi email confirmation, Github notifications...)

belugame:
he does reply to personal emails relatively quick. People tend to filter away auto-emails into folders. At least I do :)

rphlo:
I sent an email.

rphlo:
@James1345 replied 

> PyPi seem to have changed something and I can't get back into my old account. I don't know why yet, but it seems to have completely lost my email address as one that's registered to an account. 
>
> I'm contacting PyPi support to see if I can get back into the account

James1345:
Okay, email address verified, the builds should succeed now hopefully

rphlo:
Travis build went through but show `Skipping a deployment with the pypi provider because this is not a tagged commit`

James1345:
Good spot. Need to force it to try a pypi deployment somehow...

belugame:
@James1345  any update on this?

luk156:
ping

James1345:
next release should force a deployment

belugame:
just tagged master as 3.3.1, hopefully already on its way to pypi

belugame:
Release failed :( @James1345 

```
HTTPError: 403 Client Error: Invalid or non-existent authentication information. for url: https://upload.pypi.org/legacy/
django_rest_knox.egg-info/PKG-INFO already exists, no checkout
django_rest_knox.egg-info/SOURCES.txt already exists, no checkout
django_rest_knox.egg-info/dependency_links.txt already exists, no checkout
django_rest_knox.egg-info/requires.txt already exists, no checkout
django_rest_knox.egg-info/top_level.txt already exists, no checkout
Could not restore untracked files from stash entry
[...]

PR #110, opened 2018-08-10
Suggestion: make UserSerializer none by default

sphrak:
Currently `UserSerializer` hardcodes `first_name` and `last_name` on all `User` objects to be serialized as output upon successful authentication even though it does check for `USERNAME_FIELD` on the `User` object.

I believe this could be done a tad more generic -- especially checking for custom `USERNAME_FIELD` chances are that you are not even using the first_name and last_name fields on the User model. So I suggest therefore only return the `Token` and have `UserSerializer` be `None` by default:

```{"token": "9944b09199c62bcf9418ad846dd0e4bbdfc6ee4b9836F45E23A345"}```

I cannot see why one would always return the `USERNAME_FIELD` again in the response. We already supplied it when we requested the token to begin with. I am saying this for “most” cases, now there might be cases where you want it, so the option should stay but disabled by default.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1,5 +1,9 @@
 #Changelog
 
+## 3.x.x
+- **Breaking changes**: Successful authentication **ONLY** returns `Token` object by default now. 
+`USER_SERIALIZER` must be overridden to return custom data.
+
 ## 3.2.0
 - Introduce new setting AUTO_REFRESH for controlling if token expiry time should be extended automatically
 
--- a/docs/views.md
+++ b/docs/views.md
@@ -13,13 +13,17 @@ schemes. If you would like to use a different authentication scheme to the
 default, you can extend this class to provide your own value for
 `authentication_classes`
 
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: docs/changes.md [hunk 1]
```diff
@@ -1,5 +1,9 @@
 #Changelog
 
+## 3.x.x
```
belugame: you may set it to 3.3.0, though considering that we cannot release atm due to #107 I will not yet draft new releases unless bug fixes are made on master.

belugame:
looks good to me
some minor things to prepare a new release:
- please use 3.3.0 version number where you now have 3.x.x
- copy the release notes your wrote to the file https://github.com/James1345/django-rest-knox/blob/master/CHANGELOG.rst
- set the latest version number also here https://github.com/James1345/django-rest-knox/blob/master/setup.py#L19

sphrak:
@belugame I have prepared this branch to be merged and ready for 3.3.0. This seem to have created merge conflicts however..

PR #113, opened 2018-08-16
[DONT MERGE] feature: add json authentication view

sphrak:
First off, I am not sure why this doesnt already exist, maybe it was intentional but anyway here it is.
This is allows a `json` request with username and password to be sent to an endpoint instead of using Basic HTTP Authentication.

For now this is simply disabled by default, ive added some tests but I reckon more is required since it aims to authenticate a user -- even though it relies on the `authenticate()` from django to do it.

* [ ] This PR does not incorporate another PRs Ive sent just yet, because they are still pending an answer: https://github.com/James1345/django-rest-knox/pull/109
* [ ] docs are not updated yet

### DIFF
--- a/knox/serializers.py
+++ b/knox/serializers.py
@@ -1,4 +1,4 @@
-from django.contrib.auth import get_user_model
+from django.contrib.auth import authenticate, get_user_model
 
 from rest_framework import serializers
 
@@ -10,3 +10,43 @@ class UserSerializer(serializers.ModelSerializer):
     class Meta:
         model = User
         fields = (username_field, 'first_name', 'last_name',)
+
+
+class TokenSerializer(serializers.Serializer):
+
+    @property
+    def username_field(self):
+        try:
+            username_field = get_user_model().USERNAME_FIELD
+        except AttributeError:
+            username_field = 'username'
+        return username_field
+
+    def __init__(self, *args, **kwargs):
+        super(TokenSerializer, self).__init__(*args, **kwargs)
+        self.fields[self.username_field] = serializers.CharField()
+        self.fields['password'] = serializers.CharField(write_only=True)
+
+    def validate(self, attrs):
+
+        credentials = {
+            username_field: attrs.get(self.username_field),
+            'password': attrs.get('password')
+        }
+
+        if all(credentials.values()):
+            user = authenticate(**credentials)
+            if user:
+                if not user.is_active:
+                    msg = 'User account is disabled.'
+                    raise serializers.ValidationError(msg)
+                return {
+                    'user': user
+                }
+            else:
+                msg = 'Unable to login with provided credentials.'
+                raise serializers.ValidationError(msg)
+        else:
+            msg = 'Must include "{username_field}" and "password".'
+            msg = msg.format(username_field=self.username_field)
+            raise serializers.ValidationError(msg)
--- a/knox/views.py
+++ b/knox/views.py
@@ -7,6 +7,7 @@ from rest_framework.views import APIView
 
 from knox.auth import TokenAuthentication
 from knox.models import AuthToken
+from knox.serializers import TokenSerializer
 from knox.settings import knox_settings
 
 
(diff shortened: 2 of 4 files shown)


belugame:
Thank you for your work. Can you explain the need for it? Are there clients that do not support Basic Auth or in what way is sending the credentials via json better?
If there is big support for this addition I'm not against it, but at the moment I'm unsure how this makes django-rest-knox more useful.

sphrak:
Hello,
Thank you for replying.

`django-rest-knox` documentation does state that it is based on or is similiar to `TokenAuthentication` from the `django-rest-framework` -- which makes use of a json payload to authenticate. The `django-rest-framework-jwt` project does the same thing and supports http basic auth aswell -- whereas `django-rest-knox` just support http basic auth. Coming from both of those frameworks, I was surprised to
learn that `django-rest-knox` only uses http basic auth which I have come to expect with this `batteries included` concept of the Python world.

However in terms of the actual difference between http basic auth via header vs. doing it with a json payload -- there are no differences in terms of security or performance, heck one could argue that http basic auth is the most widely supported one in terms of libraries.

So my main motivation for this PR was more or less just because other frameworks support it and I wanted to even the gap between those.

### Reference documentation
* [DRF -- TokenAuthentication](https://www.django-rest-framework.org/api-guide/authentication/#generating-tokens)
* [DRF-JWT -- Usage](https://getblimp.github.io/django-rest-framework-jwt/#Usage)

sphrak:
I understand your reasoning. I do use this implementation though in my own project. Though I have to maintain a homebrew version of this project to accomplish that currently, but its okay :)

belugame:
You cannot use basic auth in your project or would it be significantly harder to use?

sphrak:
Sorry, I cannot use basic auth because it would mean breaking my own API which other services depend on. I recently switched to drf-knox since drf-jwt seem abandoned and have a bunch of glaring issues. So its less work to maintain a homebrew version with this feature, than to actually implement basic auth and deprecate an api etc..

johnraz:
I may be missing a huge point here, and I apologize if that's the case but I'm going to ask anyway because I'm puzzled:
How does this differ from the `LoginView` override recommended in the [doc](http://james1345.github.io/django-rest-knox/auth/#global-usage-on-all-views)?
Django Rest Framework will handle JSON if you specify `Content-Type: application/json`.

Cheers 😄

sphrak:
@johnraz afaik the `LoginView` still requires base64 encoded strings to be sent though in a json format no? I might be in the wrong here too. This PR allows a payload `{"username": "user", "password": "secrets"}` to be sent to the endpoint. Or am I missing something?

johnraz:
I'm currently building a new project and discovering knox and with the above mentioned `LoginView` override I can definitely login with the following request:

```
POST /api/auth/login/ HTTP/1.1
Host: localhost:8000
Content-Type: application/json
Cache-Control: no-cache

{
"username":"foo",
"password":"bar"
}
```

And I get back

```
{
    "user": {
        "username": "foo",
        "first_name": "John",
        "last_name": "Doe"
    },
    "token": "sometoken"
}
```

sphrak:
@johnraz thats awesome but I wonder what happens if you have a custom user object? In my case im not using `username`, im just using `email`.

johnraz:
I believe you should be able to customize the LoginView override to suits your custom user model.

It's using the `AuthTokenSerializer` from DJRF which in turns uses the `authenticate` method from Django itself.

I never really played that much with custom user model but I think it should work, it may even work "as-is", best chance is to give it a shot 😉

sphrak:
@johnraz alright I am gonna test that on my end and get back to you and if that works there is no need for this PR. Just to be clear tho -- when you say just specify `Content-Type: application/json` you mean on the incoming request? and then DRF will handle it appropriately as that content type?

johnraz:
@sphrak yes, so in the example I gave above:

```
POST /api/auth/login/ HTTP/1.1
Host: localhost:8000
Content-Type: application/json
Cache-Control: no-cache

{
"username":"foo",
"password":"bar"
}
```

The important bit is `Content-Type: application/json`.

Another very important bit is to override the `LoginView` ofc.

Also you need to make sure that DJRF has the [JSON parser active](http://www.django-rest-framework.org/api-guide/parsers/#formparser) but I guess it's already the case.

From there it should work, let me know what comes out of your testing.

johnraz:
@sphrak any news? I'm curious if I missed a spot here or not 😄

sphrak:
@johnraz I had forgot about the method you brought up, when I tested this library out some months ago I did exactly that. But I will have to re-visit this once my other PR is merged so I can start to use this library.

johnraz:
@sphrak: I think you should close this and re-open if it doesn't work, I'm really using it in my new project and it works /cc @belugame

belugame:
@sphrak

PR #11, opened 2016-01-30
Send out signals upon login and logout

blaklites:
This pull requests enables the sending of user_logged_in and user_logged_out signals upon login and logout.

@James1345 can this be merged?

PR #114, opened 2018-08-18
Remove settings from test projects; reload modules for tests

belugame:
alternative fix for #11 using the same workaround as d-r-f.
See commit message for more information.

### DIFF
--- a/knox/auth.py
+++ b/knox/auth.py
@@ -31,7 +31,7 @@ class TokenAuthentication(BaseAuthentication):
     authentication scheme to cope with the fact that Tokens are not stored
     in plaintext in the database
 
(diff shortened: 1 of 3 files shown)


belugame:
@rphlo  what do you think of this? It is not prettier but cleans the need of changing core code for test's sake. d-r-f from what I see is using this workaround too.

sphrak:
@belugame I like this approach, like you say it doesnt necessarily make it more pretty but it does do away with the need to changing core code for the sake for the test. 

If you choose to merge it I will update my PR https://github.com/James1345/django-rest-knox/pull/109 to reflect these changes.

PR #118, opened 2018-09-26
 Add all metadata for django 2.1 

belugame:
for building it on django 2.1 and python 3.7 to ensure there are no problems. and if to fix them


belugame:
the travis build fails cause of a bug in restframework https://github.com/encode/django-rest-framework/issues/6203

James1345:
THe fix is to specify markdown version < 3.0 (solved as part of the signals
feature)

Issue #120, opened 2018-10-03
Make MIN_REFRESH_INTERVAL a configurable settings


johnraz:
I'd like to change and make the following a configurable settings:

https://github.com/James1345/django-rest-knox/blob/370b517a20429e1cccd7b8f8bbcce10c93b766e7/knox/settings.py#L40

@rphlo, @belugame  what's your opinion as you wrote / reviewed this part?

belugame:
sure, go for it :+1:

PR #121, opened 2018-10-03
Make MIN_REFRESH_INTERVAL configurable

johnraz:
Moved the MIN_REFRESH_INTERVAL value from constants to settings.

Addresses #120


johnraz:
Thanks for the merge, note added in #125

PR #125, opened 2018-10-05
markdown all the things & release note fix

### LINKED ISSUE #124: Use only markdown
I think it's a bit sad to have to switch between RST and MD syntax in the changelog / readme and the doc.

Would you mind if I rewrite everything in markdown instead?

johnraz:
Fix #124 and address a request from #121


belugame:
i guess the build fails as README.rst is still referenced in setup.py and build_docs.sh

johnraz:
Waiting for the build but I'm confident this should be ok.

johnraz:
And ofc it failed, looking at it 😄

johnraz:
Ok I'm puzzled here, I'm going to need some help to understand why the README.md file is not available only in Python 3.4...

@belugame any idea?

excerpt from [travis logs](https://api.travis-ci.org/v3/job/439402670/log.txt):
```
cmdargs: '/home/travis/build/James1345/django-rest-knox/.tox/py34-django20/bin/python -m pip install --exists-action w /home/travis/build/James1345/django-rest-knox/.tox/.tmp/package/1/django-rest-knox-3.3.1.zip'

Looking in indexes: https://pypi.python.org/simple/
Processing ./.tox/.tmp/package/1/django-rest-knox-3.3.1.zip
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-qdfrlrpo/setup.py", line 10, in <module>
        with open(path.join(here, 'README.md'), encoding='utf-8') as f:
      File "/home/travis/build/James1345/django-rest-knox/.tox/py34-django20/lib/python3.4/codecs.py", line 891, in open
        file = builtins.open(filename, mode, buffering)
    FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pip-req-build-qdfrlrpo/README.md'
```

johnraz:
@belugame I nailed it 🎉

belugame:
great job :+1: 

could you please add a short comment to the Manifest.in that we can delete it once we drop python 3.4 support.

johnraz:
I would argue that we could keep an explicit `MANIFEST.in` file as it let us define explicitly what should be packed in the pypi release.

This also guarantee no more surprises bound to the default behaviour of a Python version (they could change their mind in coming versions...).

Issue #138, opened 2018-10-24
Push to pypi only once per build


johnraz:
The test matrix is trying to deploy to pypi for every version we have.

pypi legitimately errors because we try to upload multiple time the same file.

We should find a way to tell travis to only push to pypi once.

belugame:
https://docs.travis-ci.com/user/build-stages/matrix-expansion/

this looks exactly like what we want. it tests a matrix but only deploys one item of it.

belugame:
this also looks like it fits us with using "on:" 

https://stackoverflow.com/a/41543235/640916

PR #139, opened 2018-10-24
Limit deploy-to-pypi to single python version

belugame:
Fix for #138

We want to avoid that each item in the test matrix triggers a deployment
to pypi. This limits the deploy job to when the python version is 3.7,
hence it should only run once.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ deploy:
   on:
     tags: true
     repo: James1345/django-rest-knox
+    python: "3.7"
   only:
     - master
   distributions: "sdist bdist_wheel"


belugame:
@johnraz @James1345  what do you think? I m not familiar with travis config. But reading this https://docs.travis-ci.com/user/deployment/#conditional-releases-with-on seems promising.
I wouldnt know how we could test it easily. On the other hand we know we cannot deploy twice anway, so it cant make the situation worse.

johnraz:
Hey @belugame sorry for the delay here.

I googled a bit as I'm not used to travis either and found this: https://docs.travis-ci.com/user/build-stages from this discussion https://github.com/travis-ci/travis-ci/issues/929.

I believe your solution is good enough as a temporary fix but it would be safer to only deploy once the all test matrix passes.

What do you think ?

belugame:
Yes, that sounds like the proper way of doing it!

PR #145, opened 2018-12-09
run renew token only if the token expires

MoisesRT:
Hi, I got this error:

> delta = (new_expiry - current_expiry).total_seconds() TypeError: unsupported operand type(s) for -: 'datetime.datetime' and 'NoneType'

when I had the AUTO_REFRESH setting turned on and I logged in with a token that has no expiration time.

did a very simple bugfix


belugame:
Thank you :)

Issue #144, opened 2018-11-30
Add Remember me option


rammanoj:
I am not sure if this feature is already implemented. I am trying to set a time limit of 1 day for normal logins and for those who tick remember me, I want to set it to 1 month. In settings.py I declared globally 

` REST_KNOX = {
    'TOKEN_TTL': timedelta(hours=24),
}`

But I want to delcare it in my LoginView. I tried sending the `token_ttl` value to the parent pose method in form of kwargs. But It wasn't a successful attempt. So, asking here is there any way to implement it ??

belugame:
I would try following. Override the LoginView.post in a new class in your code, call super() to get the token and then based on request.POST.get("remember_me") you can update the expiry of the token.
Then return the Response unchanged.

rammanoj:
Thanks, @belugame, for the reply.  Is there any way to get the current AuthToken (when the user has logged in multiple systems) in serializers??.

rammanoj:
@belugame how do I get the Authtoken modal object using the token (as the token is hashed, and which field of the table should I query ??)

johnraz:
@rammanoj you can find the object by doing something like this (I tried it in a django shell and it worked): 

```
from knox.settings import CONSTANTS
user = User.objects.first()
token = AuthToken.objects.create(user=user)
AuthToken.objects.get(user=user, token_key=token[:CONSTANTS.TOKEN_KEY_LENGTH])
```

I found the answer by looking at the model manager's create method:

https://github.com/James1345/django-rest-knox/blob/517b6aad3db80684b6b56f6f8c4647e639a60731/knox/models.py#L20-L24

The problem with this approach is that you'll need to query the DB one more time to fetch back the object...

We could improve the `LoginView`'s code and have a hook method for the token creation:

https://github.com/James1345/django-rest-knox/blob/517b6aad3db80684b6b56f6f8c4647e639a60731/knox/views.py#L28

should be move to a `create_token()` method on the `LoginView` class to look like:

```
self.create_token(request)
```

That way one could easily extend the `LoginView` class and override the `create_token` method with custom logic.

But I'm not sure we want to add that kind of complexity to the `LoginView`...

belugame:
I m all for making the view more flexible, cause I remember this has been a limit for more than one user by now :)

PR #146, opened 2018-12-17
Make the LoginView more flexible

xrmx:
We need the following changes to reuse LoginView default post method instead of overriding it:
- make the context passed to the UserSerializer configurable as our serializer requires more data
- make the TTL of the AuthToken configurable instead of the default as we want to use a different TTL depending on the authentication class used

While at it create the context only if needed.

### DIFF
--- a/docs/changes.md
+++ b/docs/changes.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.5.0
+
+- The context, token TTL and tokens per user settings in `LoginView` are now dynamic
+
 ## 3.4.0
 Our release cycle was broken since 3.1.5, hence you can not find the previous releases on pypi. We now fixed the problem.
 
--- a/knox/views.py
+++ b/knox/views.py
@@ -29,11 +29,11 @@ class LoginView(APIView):
         user_logged_in.send(sender=request.user.__class__,
                             request=request, user=request.user)
         UserSerializer = knox_settings.USER_SERIALIZER
-        context = {'request': self.request, 'format': self.format_kwarg, 'view': self}
         if UserSerializer is None:
             return Response(
                 {'token': token}
             )
+        context = {'request': self.request, 'format': self.format_kwarg, 'view': self}
         return Response({
             'user': UserSerializer(request.user, context=context).data,
             'token': token,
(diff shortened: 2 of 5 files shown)


knox/views.py
```diff
@@ -16,6 +16,12 @@ class LoginView(APIView):
     authentication_classes = api_settings.DEFAULT_AUTHENTICATION_CLASSES
     permission_classes = (IsAuthenticated,)
 
+    def get_context(self):
+        return {'request': self.request, 'format': self.format_kwarg, 'view': self}
+
+    def get_token_ttl(self):
```
MoisesRT: why this function is needed?

knox/views.py
```diff
@@ -25,15 +31,16 @@ def post(self, request, format=None):
                     {"error": "Maximum amount of tokens allowed per user exceeded."},
                     status=status.HTTP_403_FORBIDDEN
                 )
-        token = AuthToken.objects.create(request.user)
+        token_ttl = self.get_token_ttl()
+        token = AuthToken.objects.create(request.user, token_ttl)
```
MoisesRT: it's already doing that by default:
https://github.com/James1345/django-rest-knox/blob/master/knox/models.py#L12

xrmx:
As written in the cover letter I want to be able to set a TTL depending on
the request

Il giorno lun 17 dic 2018, 20:15 Moises Rezonzew <notifications@github.com>
ha scritto:

belugame:
Thank you for your contribution, looks good to me.
Are you happy to add a new minor version for this with a short description to the changelog and the docs?

xrmx:
Sure will update the  PR tomorrow

johnraz:
Cool 👍 - Let's make sure to close #144 when merged.

xrmx:
I've updated the PR adding docs and while at it made the token per user limit configurable too.

docs/changes.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.4.1
```
johnraz: Maybe bumping to 3.5.0 would make sense as we are adding a feature?
Also make sure to bump the version in `setup.py` too.

xrmx: Makes sense, pushed updated branch.

belugame:
great, thanks!

CHANGELOG.md is still missing though I find it rather bothersome to maintain 3 places of change logs (docs, root, and github release description). should we get rid of at least one?

xrmx:
@belugame fixed!

johnraz:
Sweet! @belugame  I created #148 to keep track of the changelog matter.

PR #149, opened 2018-12-27
Make the user serializer configurable for LoginView

xrmx:
So that's easily overridable when hosting more than one instance
of LoginView in the same project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+3.5.1
+=====
+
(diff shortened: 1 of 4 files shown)


belugame:
oh, merged it a bit early. strictly speaking this is not a bugfix, so it should be 3.6.0
and we need to amend the version number also in setup.py. then i can release it

xrmx:
Thanks for merging, I'm not in a hurry to have a release with this included but still created a PR for 3.6.0 :)

PR #152, opened 2019-01-07
fixes #131 -- return expiry date on successful login

### LINKED ISSUE #131: Return expiring date after Login
Is there a reason why only the user is returned along with the token and not also the expiring date or the expiring time-delta?

It could be very helpful to have this information to validate tokens in the frontend before using them.

sphrak:
Hello,
I have looked at suggestion #131 and concluded that its probably most wise to solve it by two PR's instead of one.

cc @johnraz talks about an endpoint which you can submit a token to and get the expiration from, while this is somewhat related I think it should be a seperate PR (but correct me if you think otherwise and ill include it here).

I have intentionally made this setting disabled by default to not break existing installations, but
setting it to true will give the desired output.

Still missing:
* [x] documentation
* [x] changelog 4.0.0

Please give me feedback :octocat: 

Kindly


docs/settings.md
```diff
@@ -55,6 +56,11 @@ Setting the TOKEN_TTL to `None` will create tokens that never expire.
 Warning: setting a 0 or negative timedelta will create tokens that instantly expire,
 the system will not prevent you setting this.
 
+## TOKEN_TTL_VISIBILITY
```
johnraz: I'm not 100% sure about the name, as usual naming things is hard...
I feel that we should add a clue that this is related to the login view only.

`LOGIN_TOKEN_TTL_VISIBILITY`
`LOGIN_SHOW_TOKEN_TTL`
`SHOW_TOKEN_TTL_AT_LOGIN` ...

knox/views.py
```diff
@@ -38,20 +41,41 @@ def post(self, request, format=None):
...
+                    'token': token,
+                    'token_expiry': token_ttl,
+                }
+            )
+        return Response(
+            {
+                'user': UserSerializer(request.user, context=context).data,
+                'token': token
+            }
+        )
```
johnraz: - There is no need to check specifically for `True` and `False` any value that can be casted to boolean should be fine.

- The logic could be simplified and made more readable by doing:
```
if UserSerializer is None:
    resp = {...}
else:
    resp = {...}

if show_token_ttl:
    resp["token_expiry"] = token_ttl

return resp

Comment (sphrak) [reply knox/views.py]: All good points and yes, it did feel a tad ugly to do those if-statements @johnraz, I will fix it asap :)

File: docs/settings.md (hunk 1 above)
Comment (sphrak) [reply docs/settings.md]: `ENABLE_TOKEN_EXPIRY`? Yeah naming is hard. Lets think about it some, maybe @belugame have some input? :-)

File: knox/views.py (hunk 2 above)
Comment (sphrak) [reply knox/views.py]: @johnraz its done :) but hey, my naming tho, `data`? haha

File: docs/settings.md (hunk 1 above)
Comment (johnraz) [reply docs/settings.md]: Mmmm not a  fan of `ENABLE_TOKEN_EXPIRY` it sounds like it would enable / disable the token expiry feature like "a token would never expire when falsy".

`LOGIN_TOKEN_TTL` ?

Also the more I think about the name, the more I'm wondering if we need a setting at all.
This is not backward incompatible as it will just add a key to the payload, not change any of the existing keys + there is no extra performance cost when returning it.

Maybe we should just drop the setting?

File: tests/tests.py [hunk 3]
```diff
@@ -321,3 +328,72 @@ def test_invalid_prefix_return_401(self):
...
+                True
+            )
+            self.assertEqual(
+                token_ttl_visibility_user_serializer["USER_SERIALIZER"],
+                UserSerializer
+            )
+            self.assertEqual(response.status_code, 200)
+            self.assertIn('user', response.data)
+            self.assertIn('token', response.data)
+            self.assertIn('token_expiry', response.data)
```
johnraz: We should also check for the `token_expiry` value and its formatting.

knox/views.py
johnraz: `data` is great 👍

knox/views.py
```diff
@@ -38,20 +41,29 @@ def post(self, request, format=None):
...
                 'token': token
-            })
-        context = self.get_context()
-        return Response({
-            'user': UserSerializer(request.user, context=context).data,
-            'token': token,
-        })
+            }
+        if show_token_ttl:
+            data['token_expiry'] = token_ttl
```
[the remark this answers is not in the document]
sphrak: oh nooo, you are right I completely misread that part. I will scamble a fix.

docs/settings.md
johnraz: I do agree that keeping consistency in default values and behaviours is important.

On the other hand, I think we can also safely say that knowing when and if a token expires is not solely linked to the auto refresh feature:

- The token expiry can be used when auto refresh is off, eg: to warn a user that he will be disconnected in x minutes.

- You may also want to let your user know that he needs to explicitly logout when his token never expires.

To make a good argument in favor of a configurable token expiry visibility we would need to find a case against exposing the token expiry and I struggle finding one, any thoughts?

belugame:
I don't feel a need for a setting for this as I can't see a big risk of it breaking existing installations. People don't usually check on the number of keys in the response or something. Could you elaborate in what kind of frontend setup it would break?

And as we would release a new version anyway, it gives people the choice to stay on the old version as long as they like before adapting, which I would think would be a minor task to do.

knox/views.py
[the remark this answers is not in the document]
sphrak: @belugame sounds good, I am in favor of re-writing those parts to comply with "what is expected" since it is afterall a django project. @johnraz would you agree or do you want another direction?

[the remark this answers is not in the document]
johnraz: > What is the main reasoning to we not simply do the what django does and return the model instance? It gives alot more flexibility, no need to even return a tuple of the "instance + a duplicate part of the instance"?

The token, the string, is not stored as is in the `AuthToken`  but it's hashed with a salt.
That's why we need the tuple. 

> Adding new methods to circumvent old oddities is counterproductive I feel.
I do agree, so I would say that we should implement the "tuple" like behavior in the existing create method. Using create without returning the token is non-sense anyway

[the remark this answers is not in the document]
sphrak: @johnraz Yes, I just realized that, sorry for the noise. I could just overwrite the `instance.token_key` with the raw `token` before returning it thus getting best of both worlds (simplicity and the functionality we need). Ill type something up, ill have it ready in a bit.

I agree with you and @belugame I cannot find a good reason to keep the setting for this, and just always include it unless the expiration is set to `None` then obviously we dont return it.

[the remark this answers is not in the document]
johnraz: I would always return the `token_ttl`, if it's `None` it will be `null` in the payload and that's fine. Otherwise frontend libraries would need to check for existence etc.

I wouldn't erase the value of `instance.token_key` because it will make things confusing and implicit, to quote the zen of Python "Explicit is better than implicit" and I think the most explicit solution is to return a tuple. It changes the default return value of a create method which is not ideal but at the same time it is not hidden and one would easily spot that the method returned a tuple by reading our code:

Override the `token_key`
```
instance = AuthToken.objects.create() <-- no sign of an overridden token_key here ; implicit behavior :/
```

Tuple
```
instance, token = AuthToken.objects.create() <-- explicitly assign the token which is more readable.
```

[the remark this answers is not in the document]
sphrak: @johnraz you are absolutly right, I am moving to quickly. I ended up implementing what you said. Tossed the setting and made the create method return the instance and token :smiley_cat:

knox/views.py
```diff
@@ -39,19 +39,26 @@ def post(self, request, format=None):
...
+                ).data,
                 'token': token
-            })
-        context = self.get_context()
-        return Response({
-            'user': UserSerializer(request.user, context=context).data,
-            'token': token,
-        })
+            }
+        data['expiry'] = instance.expires
```
sphrak: My thinking was keep it dry here

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+4.0.0
+=====
+
+- Successful login now always return a `expiry` field for when the token expires.
```
johnraz: I would mention the breaking change and action to be taken in order to upgrade.

knox/views.py
johnraz: works for me and I like the `expiry` name 🤓

tests/tests.py
```diff
@@ -164,14 +167,15 @@ def test_expired_tokens_deleted(self):
...
-        token = AuthToken.objects.create(self.user)
+        instance, token = AuthToken.objects.create(self.user)
         rf = APIRequestFactory()
         request = rf.get('/')
         request.META = {'HTTP_AUTHORIZATION': 'Token {}'.format(token)}
         (self.user, auth_token) = TokenAuthentication().authenticate(request)
         self.assertEqual(
             token[:CONSTANTS.TOKEN_KEY_LENGTH],
-            auth_token.token_key)
+            auth_token.token_key
```
johnraz: Nitpick alert: I would expect a comma at the end of the line.

sphrak: Just to be clear, like this?
```
        self.assertEqual(
            token[:CONSTANTS.TOKEN_KEY_LENGTH],
            auth_token.token_key,
        )
```

johnraz: Exactly :-)

CHANGELOG.md
sphrak: @johnraz do you think I need to mention more? afaik unless you rely on the create() method directly, the upgrade path should be quite simple no?

johnraz: Yes I think that anybody could be using the AuthToken model with custom view to handle various things (like the refresh-token view we are missing).

sphrak:
@johnraz my pleasure mate, thanks for all the help with code review :)

CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
+4.0.0
+=====
+
+**BREAKING** This is a major release version because it
+breaks the existing API.
+Changes has been made to the `create()` method on the `AuthToken` model.
```
belugame: minor typo: *have* been made

CHANGELOG.md
```diff
@@ -1,3 +1,18 @@
...
+breaks the existing API.
+Changes has been made to the `create()` method on the `AuthToken` model. 
+It now returns the model instance and the raw `token` instead
+of just the `token` to allow the `expiry` field to be included in the
+success response.
+
+Depending on your usage you might have to adjust your code
+to fit these new changes.
+
+- Successful login now always return a `expiry` field for when the token expires.
```
belugame: minor typo: *returns* a

knox/views.py
belugame: could we turn it around to save some lines?

like this (no need to copy it literally):

```
data = {"expiry": instance.expires, "token": token}
if UserSerializer is not None:
   data["user"] = ...
```

what i personally dont like is that we then have 2 names for the same thing, which is always confusing and frustrating for someone learning: `expiry` on the response vs. `expires` on the model.
I also like `expiry` better for the reasons someone else has mentioned. and we could in the same instant change the model field name too as we make this a major release already anyway.

what do you think? :)

johnraz: I'm all in for the field rename, I don't like it.
Putting a DB migration on the token expiry should again be properly advertised in the changelog, I'm not sure how alter table for a rename works in other DB engine but I know postgres would take a full lock on a rename operation...
You definitely would want to take appropriate action to make sur that it's at least as fast as possible :-)

sphrak: Hello guys,
I fully agree with you, we should change the fieldname from `expires` to `expiry`, its more clean. Latest commit have these changes, however @johnraz could you elaborate what you meant with "making sure its as fast as possible"? Afaik this is only a field rename, ought to be pretty fast no?

Relevant docs: 
- https://docs.djangoproject.com/en/2.1/ref/migration-operations/#renamefield
- https://docs.djangoproject.com/en/2.1/_modules/django/db/migrations/operations/fields/#RenameField

sphrak:
woho :tada: thanks @belugame and @johnraz -- pleasure as always :)

knox/views.py
johnraz: @sphrak in most scenarios it should be fast enough indeed, with a very large token table though it could lock the table for a fairly annoying amount of time but I'm probably overthinking this.

I was thinking about investigating a solution like copying the data to the new column name and then dropping the old one instead of renaming, to avoid the lock. But I have not thought about this much longer and can't say if it would bring any improvements over the rename, I'm glad this is now merged btw 🎉

sphrak: @johnraz I'm not entirely sure but I would assume in the case of postgres that such a "migration" like renaming a column/field is translated directly into sql syntax `RENAME COLUMN column_name TO new_column_name;` which ought to be much faster in postgres C code than running a data migration in python -- more so at least under the conditions you where concerned about.

but yes, time will tell if we made a wopsy but I am also happy its merged now. Again thank you both for the code reviews, I really enjoy working with you guys. :grinning: 

On a different note (working on the refresh endpoint PR), you dont happen to know if I can reload the `urls.py` via `reload_module` in testing? I cannot get it to work. I am trying to implement a conditional url setting so that we only expose the refresh endpoint if the user has `AUTO_REFRESH` and `TOKEN_TTL` set to not `None` otherwise it wont make sense to have it exposed.

We can continue the discussion in #156 :)

belugame:
@sphrak @johnraz Thank you a lot for your continued effort and contributions!
I will make the 4.0 release on Monday. Don't wanna surprise someone who uses pyup on a Friday.

knox/views.py
johnraz: I didn't mean a python data migration as it's obviously slower, you can run arbitrary SQL queries in Django migration. I was more thinking on how to compose those SQL queries to achieve a non blocking solution.

I'm not sure about `reload_module` I haven't been touching the tests for a while maybe @belugame knows better but yea let's keep discussing on the other ticket 😉

PR #155, opened 2019-01-08
fixes #154 -- adds isort

### LINKED ISSUE #154: Add isort to project
Hello,
I see this project does not use [isort](https://github.com/timothycrosley/isort). Maybe we could include it to improve readability of import statements? I am willing to do it if you think its a good idea.

Kindly,

sphrak:
Hello, 
Adds isort to testenv and sorts all imports. Closes https://github.com/James1345/django-rest-knox/issues/154 :cat2:

We might want to add some extra rules, personally I run these in my projects:
```
[isort]
combine_as_imports = true
default_section = THIRDPARTY
include_trailing_comma = true
known_first_party = django
line_length = 79
multi_line_output = 5
not_skip = __init__.py
```

Also, this does not test the imports in `tests/`. Let me know if that feels like something you might want.

Cheers mates

### DIFF
--- a/knox/auth.py
+++ b/knox/auth.py
@@ -6,14 +6,12 @@ except ImportError:
 
 import binascii
 
-from django.utils.translation import ugettext_lazy as _
-from django.utils import timezone
 from django.contrib.auth import get_user_model
-
+from django.utils import timezone
+from django.utils.translation import ugettext_lazy as _
 from rest_framework import exceptions
 from rest_framework.authentication import (
-    BaseAuthentication,
-    get_authorization_header
+    BaseAuthentication, get_authorization_header,
 )
 
 from knox.crypto import hash_token
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,0 +1,7 @@
+[settings]
+combine_as_imports = true
+default_section = THIRDPARTY
+include_trailing_comma = true
+known_first_party = knox
+multi_line_output = 5
+not_skip = __init__.py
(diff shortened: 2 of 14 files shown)


knox/auth.py
```diff
@@ -6,20 +6,16 @@ def compare_digest(a, b):
...
-
+from django.utils import timezone
+from django.utils.translation import ugettext_lazy as _
 from knox.crypto import hash_token
 from knox.models import AuthToken
 from knox.settings import CONSTANTS, knox_settings
 from knox.signals import token_expired
+from rest_framework import exceptions
+from rest_framework.authentication import (BaseAuthentication,
+                                           get_authorization_header)
```
johnraz: Can we configure isort so that it gives:
```
from rest_framework.authentication import (	
    BaseAuthentication,	
    get_authorization_header,
)
```

belugame:
awesome, 2 quick remarks:

1) lets do it also in tests, there is no reason not to
2) I would prefer to have knox imports below those libaries knox uses. To show which builds on what and making it the last instance to rule over others. So a knox import should be after a restframework import in my opinion.

knox/auth.py
sphrak: Yes my friend, consider it done :)

sphrak:
@belugame we can configure `knox` to be known first party as shown now in this current commit. Is that what you are looking for?

belugame:
yes, great :) :+1:

Issue #157, opened 2019-01-14
DRF request throttling ignored by LoginView


filipweidemann:
I just stumbled over an issue with the LoginView. Specifying global throttling classes and rates doesn't seem to stop the LoginView from accepting and processing requests anyway.
Even when creating custom throttling classes the same result happens.

Is this intended behaviour? Typically it shouldn't happen that a login view is not throttled..

johnraz:
Hello @filipweidemann,

This is definitely not the intended behaviour.

I'm a bit puzzled that the global throttling is not triggered as, at first glance, I don't see anything special in the `LoginView` that could avoid the throttling to work.

I would like to make sure that the issue lies with this library and is not a configuration issue.

Is the throttling working with other views than the login view?

Would you mind sharing some code to help reproduce the issue?

Code samples would be good but If you could setup a project demonstrating the issue or even better provide a PR with a failing test case, that would help a ton.

belugame:
I would also be surprised if knox affects drf throttling in any way. @filipweidemann the issue does not happen to you on drf's login view?

filipweidemann:
First of all, thanks for the quick responses.

@johnraz 
I already did the same, checked the `LoginView`, saw nothing out of the ordinary, so I started to limit the requests for anonymous users to `1/min` for easy testing. The result was that every single `APIView` was restricted, except the `LoginView`.

I created a sample project with just two views, one is the `LoginView`, the second one is a random `APIView` that just returns some content. 

@belugame 
I didn't know that DRF had built-in login views, so I can't say if that works or not.

I created a sample project with which you can easily reproduce the issue, have a look: https://github.com/filipweidemann/knox-throttling

johnraz:
Allright, thanks for the example project @filipweidemann it helped me a ton to get started and test this out.

I can confirm this is not related to knox.

A bit of context:
First we need to make it clear that there is authenticated and anonymous throttling playing a confusing game here...

In your project only anonymous throttling is enabled.
If you'd like to block someone from frenetically call the login view and get new tokens like crazy, you'd need to enable the authenticated throttling as well.

So I decided to ignore the above and consider only the case where you'd want to block someone that has no valid credentials and is trying to brute-force your login view.

At that stage, only the authentication middleware is relevant and not the login view at all.

How I came to this conclusion:

- hit the login view with credentials --> repeating the request doesn't trigger the throttle and it's expected behavior as we are dealing with anonymous user throttling only.

[...]

PR #158, opened 2019-01-14
fixes #156 -- provide token refresh endpoint

### LINKED ISSUE #156: [discussion] Provide endpoint for checking token validity
As discussed in #131 @johnraz suggested that we also provide an endpoint for refreshing or increasing the expiration date by simply hitting the new endpoint and the tokens expiration date is increased.

I am thinking that we provide a new url `/refresh/` that just requires an authenticated `PATCH` request to the endpoint which returns the new expiry for the token: `{"expiry": "2019-01-11T07:54:34.504745"}`

There should also be some control over how often this can happen. Not sure what to do here though.

I am willing to take this on me, but I would like to hear what you think before I start :)

sphrak:
Hi,
First draft of whats discussed in #156, I ended up going for a simpler approach than integrating drf throttling due to the points laid out by @johnraz. However as suggested by #157 we should prolly take a look at that aswell since at least I would expect drf throttling to also apply to our views. (different issue tho).

As discussed these endpoints are optional, and they are not exposed by default unless `TOKEN_TTL` and `AUTO_REFRESH` are set. Where `AUTO_REFRESH` is `False` by default, thus off. 

Let me know what you think

### DIFF
--- a/docs/urls.md
+++ b/docs/urls.md
@@ -1,4 +1,4 @@
-#URLS `knox.urls`
+# URLS `knox.urls`
 Knox provides a url config ready with its three default views routed.
 
 This can easily be included in your url config:
@@ -17,12 +17,14 @@ The views would then acessible as:
 
 - `/api/auth/login` -> `LoginView`
 - `/api/auth/logout` -> `LogoutView`
-- `/api/auth/logoutall` -> `LogoutAllView`
+- `/api/auth/logout/all` -> `LogoutAllView`
+- `/api/auth/refresh` -> `TokenRefreshView`
 
 they can also be looked up by name:
 
 ```python
 reverse('knox_login')
 reverse('knox_logout')
-reverse('knox_logoutall')
+reverse('knox_logout_all')
+reverse('knox_refresh')
 ```
--- a/docs/views.md
+++ b/docs/views.md
@@ -53,3 +53,11 @@ system and can no longer be used to authenticate.
 **Note** It is not recommended to alter the Logout views. They are designed
 specifically for token management, and to respond to Knox authentication.
 Modified forms of the class may cause unpredictable results.
+
+## TokenRefreshView
+This view accepts only a post request with an empty body.
+On a successful request, the token used to authenticate will be refreshed and its expiry will be extended by the amount set in `TOKEN_TTL`.
+
+The response body includes an expiry key with a timestamp that represents the token's new expiry. 
+
+**Note**: To enable this view set `ENABLE_REFRESH_ENDPOINT` to `True`.
--- a/knox/auth.py
+++ b/knox/auth.py
@@ -37,6 +37,9 @@ class TokenAuthentication(BaseAuthentication):
     model = AuthToken
 
     def authenticate(self, request):
+
+        # view = getattr(request, 'parser_context', None)['view']
+        view = request.parser_context['view']
         auth = get_authorization_header(request).split()
         prefix = knox_settings.AUTH_HEADER_PREFIX.encode()
 
@@ -50,10 +53,10 @@ class TokenAuthentication(BaseAuthentication):
                     'Token string should not contain spaces.')
             raise exceptions.AuthenticationFailed(msg)
 
-        user, auth_token = self.authenticate_credentials(auth[1])
+        user, auth_token = self.authenticate_credentials(auth[1], view)
         return (user, auth_token)
 
-    def authenticate_credentials(self, token):
+    def authenticate_credentials(self, token, view):
         '''
         Due to the random nature of hashing a salted value, this must inspect
         each auth_token individually to find the correct one.
@@ -72,7 +75,8 @@ class TokenAuthentication(BaseAuthentication):
             except (TypeError, binascii.Error):
                 raise exceptions.AuthenticationFailed(msg)
             if compare_digest(digest, auth_token.digest):
-                if knox_settings.AUTO_REFRESH and auth_token.expiry:
+                auto_refresh = getattr(view, 'auto_refresh', knox_settings.AUTO_REFRESH)
+                if auto_refresh and auth_token.expiry:
                     self.renew_token(auth_token)
                 return self.validate_user(auth_token)
         raise exceptions.AuthenticationFailed(msg)
--- a/knox/urls.py
+++ b/knox/urls.py
@@ -4,6 +4,7 @@ from knox import views
 
 urlpatterns = [
     url(r'login/', views.LoginView.as_view(), name='knox_login'),
-    url(r'logout/', views.LogoutView.as_view(), name='knox_logout'),
-    url(r'logoutall/', views.LogoutAllView.as_view(), name='knox_logoutall'),
+    url(r'logout/$', views.LogoutView.as_view(), name='knox_logout'),
+    url(r'logout/all/$', views.LogoutAllView.as_view(), name='knox_logout_all'),
+    url(r'refresh/$', views.TokenRefreshView.as_view(), name='knox_refresh')
 ]
--- a/knox/views.py
+++ b/knox/views.py
@@ -79,3 +79,30 @@ class LogoutAllView(APIView):
         user_logged_out.send(sender=request.user.__class__,
                              request=request, user=request.user)
         return Response(None, status=status.HTTP_204_NO_CONTENT)
+
+
+class TokenRefreshView(APIView):
+
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAuthenticated,)
+    auto_refresh = False
+
+    def post(self, request, format=None):
+
+        if not knox_settings.ENABLE_REFRESH_ENDPOINT:
+            return Response(
+                status=status.HTTP_403_FORBIDDEN
+            )
+
+        if knox_settings.TOKEN_TTL is None:
+            raise ValueError(
+                'Value of \'TOKEN_TTL\' cannot be \'None\' in this context.'
+            )
+
+        request.auth.expiry = timezone.now() + knox_settings.TOKEN_TTL
+        request.auth.save()
+
+        return Response(
+            {'expiry': request.auth.expiry},
+            status=status.HTTP_200_OK
+        )
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -1,6 +1,7 @@
 import base64
 from datetime import datetime, timedelta
 
+import pytz
 from django.contrib.auth import get_user_model
 from django.test import override_settings
 from django.utils.six.moves import reload_module
@@ -44,6 +45,13 @@ auth_header_prefix_knox["AUTH_HEADER_PREFIX"] = 'Baerer'
 
 token_no_expiration_knox = knox_settings.defaults.copy()
 token_no_expiration_knox["TOKEN_TTL"] = None
+
+refresh_endpoint_knox = knox_settings.defaults.copy()
+refresh_endpoint_knox["ENABLE_REFRESH_ENDPOINT"] = True
+
+token_ttl_is_none_knox = knox_settings.defaults.copy()
+token_ttl_is_none_knox["ENABLE_REFRESH_ENDPOINT"] = True
+token_ttl_is_none_knox["TOKEN_TTL"] = None
 
 
 class AuthTestCase(TestCase):
@@ -119,7 +127,7 @@ class AuthTestCase(TestCase):
             instance, token = AuthToken.objects.create(user=self.user)
         self.assertEqual(AuthToken.objects.count(), 10)
 
-        url = reverse('knox_logoutall')
+        url = reverse('knox_logout_all')
         self.client.credentials(HTTP_AUTHORIZATION=('Token %s' % token))
         self.client.post(url, {}, format='json')
         self.assertEqual(AuthToken.objects.count(), 0)
@@ -131,7 +139,7 @@ class AuthTestCase(TestCase):
             AuthToken.objects.create(user=self.user2)
         self.assertEqual(AuthToken.objects.count(), 20)
 
-        url = reverse('knox_logoutall')
+        url = reverse('knox_logout_all')
         self.client.credentials(HTTP_AUTHORIZATION=('Token %s' % token))
         self.client.post(url, {}, format='json')
         self.assertEqual(AuthToken.objects.count(), 10,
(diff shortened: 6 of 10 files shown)


knox/views.py
```diff
@@ -79,3 +79,33 @@ def post(self, request, format=None):
...
+        )
+
+
+class TokenRefreshAllView(APIView):
+
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAuthenticated,)
+
+    def post(self, request, format=None):
+        request.user.auth_token_set.all().update(
```
sphrak: This should probably be filtered so that we dont make expired tokens valid and non expired again.

knox/views.py
```diff
@@ -79,3 +79,33 @@ def post(self, request, format=None):
...
+            expiry=timezone.now() + knox_settings.TOKEN_TTL
+        )
+        """either we return expiry for 1 key which
+        is the same for all keys, or we return nothing
+        and set response code 204 indicating success,
+        but no data returned"""
+        return Response(
```
sphrak: I am not sure what makes most sense here. We update potentially multiple instances, do we just return `expiry` and imply its the new time for all tokens.. or do we just return 204?

docs/urls.md
```diff
@@ -18,11 +18,15 @@ The views would then acessible as:
 - `/api/auth/login` -> `LoginView`
 - `/api/auth/logout` -> `LogoutView`
 - `/api/auth/logoutall` -> `LogoutAllView`
+- `/api/auth/refresh` -> `TokenRefreshView`
+- `/api/auth/refresh/all` -> `TokenRefreshAllView`
```
johnraz: I would rename `logoutall` to `logout/all` to keep things consistent.
As usual adding a note in the changelog that this is backward incompatible and can be restored easily by creating a route manually at the project level.

docs/urls.md
```diff
@@ -18,11 +18,15 @@ The views would then acessible as:
...
 
 they can also be looked up by name:
 
 ```python
 reverse('knox_login')
 reverse('knox_logout')
 reverse('knox_logoutall')
+reverse('knox_refresh')
+reverse('knox_refresh_all')
```
johnraz: Goes with my comment above: `knox_logout_all` => `knox_logout_all`

docs/views.md
```diff
@@ -53,3 +53,14 @@ system and can no longer be used to authenticate.
 **Note** It is not recommended to alter the Logout views. They are designed
 specifically for token management, and to respond to Knox authentication.
 Modified forms of the class may cause unpredictable results.
+
+## TokenRefreshView
+Optional view for refreshing the expiration date of a single token.
+It accepts an empty `POST` request against `/refresh` to increase the `expiry` field of the
+token by the amount set in `TOKEN_TTL`.
+This endpoint is only exposed if `AUTO_REFRESH` and `TOKEN_TTL` are enabled.
```
johnraz: I would argue that the endpoint would still be useful if one want's to disable `AUTO_REFRESH` but still want to be able to manually refresh a token.

I see the point of not enabling the endpoint if `TOKEN_TTL` is `None` as no token would ever expire.
So I would rephrase this a bit and make it clear that we clearly specify that it's only disabled when `TOKEN_TTL` is `None`.

Also see my comment below about this feature.

docs/views.md
```diff
@@ -53,3 +53,14 @@ system and can no longer be used to authenticate.
...
 specifically for token management, and to respond to Knox authentication.
 Modified forms of the class may cause unpredictable results.
+
+## TokenRefreshView
+Optional view for refreshing the expiration date of a single token.
+It accepts an empty `POST` request against `/refresh` to increase the `expiry` field of the
+token by the amount set in `TOKEN_TTL`.
+This endpoint is only exposed if `AUTO_REFRESH` and `TOKEN_TTL` are enabled. 
+
+The functionality can be disabled entirely by setting `DISABLE_REFRESH_ENDPOINT=True` in `REST_KNOX` settings.
```
johnraz: We could get rid of the logic explained above by just having a `ENABLE_REFRESH_ENDPOINT` setting.
It would be easier to explain and would also make the code simpler.

The explanation would boil down to:

> If you want the refresh endpoint, just opt-in by setting `ENABLE_REFRESH_ENDPOINT` to `True`.

Simple, straight, obvious, no magic.

knox/urls.py
```diff
@@ -1,9 +1,25 @@
...
+        views.TokenRefreshView.as_view(),
+        name='knox_refresh'
+    ),
+    url(
+        r'refresh/all/$',
+        views.TokenRefreshAllView.as_view(),
+        name='knox_refresh_all'
+    ),
+]
+if knox_settings.TOKEN_TTL and knox_settings.AUTO_REFRESH:
```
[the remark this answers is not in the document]
johnraz: I guess my comments above answers this question and my opinion on this.

knox/views.py
```diff
@@ -79,3 +79,33 @@ def post(self, request, format=None):
...
         return Response(None, status=status.HTTP_204_NO_CONTENT)
+
+
+class TokenRefreshView(APIView):
+
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAuthenticated,)
+
+    def post(self, request, format=None):
+        return Response(
```
johnraz: We should do an explicit update on the expiry here:

- If we don't, in the context of my previous comments, it will not work for people with `AUTO_REFRESH=False` and wanting to manually refresh.

- this view should bypass `MIN_REFRESH_INTERVAL` as it is an explicit request for a refresh and not an implicit one triggered by `AUTO_REFRESH`.

Ideally, we should find a way to disable the behavior of `AUTO_REFRESH` in the authentication class when hitting this endpoint.

knox/views.py
johnraz: It shouldn't happen as the cleanup phase is triggered during the authentication phase for all token of the identified user:

https://github.com/James1345/django-rest-knox/blob/c1972fcddf86179680522d6af1b463b02c49b00c/knox/auth.py#L58-L70

https://github.com/James1345/django-rest-knox/blob/c1972fcddf86179680522d6af1b463b02c49b00c/knox/auth.py#L100-L115

knox/views.py
johnraz: It's probably more consistent to return the `expiry`.
I agree this boils to implying something but at the same time we just introduced the fact that the login view would return the expiry and the refresh/token view also returns the expiry.
I personally would be surprised not having the expiry returned.

johnraz:
Oh and regarding throttling, I think it's a picky subject.
I like the idea of making it easier on people to enable it but at the same time it can be easily broken.
eg: by a bad proxy configuration.
We need to make sure we think it through and document it properly, I suggest we get to the bottom of #157 and then move on and decide based on the outcome.

knox/views.py
belugame: I agree with @johnraz

docs/urls.md
belugame: I edited your comment, think you typed before and after as the same :) It's weird that github allows one to edit other people's comments I feel.

knox/urls.py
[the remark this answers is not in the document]
belugame: hm I feel there is no real gain of this flexible url setup, if I miss something let me know. We make our lifes as devs harder in terms of testing etc but do we decrease security or anything if we have this url "avaiable" despite not allowing refreshes? The view could simply return a 401.

knox/views.py
belugame: could you describe a use case where one client needs to refresh all its token? i would think they are not used in one device and therefore it does not make sense to me to extend the tokens of non-active devices.

knox/urls.py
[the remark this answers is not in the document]
sphrak: @belugame I agree, lets do that instead. Lets return 403 as per: 

> This status is similar to 401, but in this case, re-authenticating will make no difference. The access is permanently forbidden and tied to the application logic, such as insufficient rights to a resource.

knox/views.py
sphrak: @johnraz I agree with you, ill go back to the drawing board a bit to try to make this more explicity and flexible.

knox/views.py
sphrak: @belugame I cannot, but it was mentioned in the discussion thread, I know I thought it was a good idea but now I am not so sure anymore because the more I think of it, the less convinced I become. It should be up to each individual client to do this, not the responsability of one client to do it for everybody else.

@johnraz what do you think about this? Should we nuke this feature?

knox/views.py
sphrak: @johnraz I am not sure how to circumvent the auto refresh in auth without making it messy. I mean we could explicitly do it in the refresh view if `AUTO_REFRESH` is not set, but if we do that it will not bypass `MIN_REFRESH_INTERVAL` if the user happen to be using `AUTO_REFRESH` also.. which is kinda unexpected behaivor if you ask me..

knox/urls.py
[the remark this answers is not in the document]
johnraz: Agreed with the above + 403 👍

knox/views.py
johnraz: Yea I realize as well that this is more than a niche feature. I can't even remember the precise use-case I had in mind, I think I was confused by the logout all use-case which makes perfect sense... Nuke! 💥

knox/views.py
johnraz: So the auto-refresh flow today is:
- hit a view with a token -->  if `MIN_REFRESH_INTERVAL` allows it, update expiry and hit the DB with a write.
- hit a view with a token -->  if `MIN_REFRESH_INTERVAL` denies it, don't update expiry and don't write to the DB.

What we'd ideally want for the refresh view (whether or not auto-refresh is active) is:
- hit the refresh view with a token --> refresh the token in every case by doing a single write to the DB

If auto-refresh is off, there is no issue.

If auto-refresh is on, hitting the refresh view will do this:
- if `MIN_REFRESH_INTERVAL` is expired, the auto-refresh will update the expiry and write to the DB at authentication time, next the view will do it again, resulting in 2 writes for nothing.
- if `MIN_REFRESH_INTERVAL` is not expired, only the view will write to the DB, 1 write everybody is happy.

The problem we need to solve is to avoid to have 2 consecutive useless writes to the DB...
Sadly I don't think we have access to the view or anything useful to identify the refresh view in the authenticate method (where `AUTO_REFRESH` happens)...

I would say, let the 2 DB write be and add a note in the doc that if you use `AUTO_REFRESH` and the refresh view together, a double write could happen.

johnraz: Taking the above statement back.
The view can be accessed in the authenticate method by doing:
```
request.parser_context['view']
```

This means we could simply add a property to the class like:
```
class RefreshView():
    disable_auto_refresh=True
```

And then check for that in the `authenticate` / `authenticate_credentials` method of the `TokenAuthentication` to disable auto-refresh "per view"

sphrak: @johnraz, wonderful research, thank you. I have been experimenting with this approach, but it seems to be not as trivial as it initially seemed. 

If we reference `AUTO_REFRESH` value in the authenticate/_credentials() methods it needs to be accessible on every view which kinda sucks because then we need to add it to each view or do some weird complex thing of trying to validate who or which part of the code is calling us -- which I dont think we should do since this is a library. 

I dont like that so maybe we should make a `KnoxAPIView` class that extends from `APIView` with the only difference this attribute `AUTO_REFRESH` is added as a field attribute like `auto_refresh` or so.

But I am not sure if its a good approach, though I would like to be able to do away with things like `if knox_settings.TOKEN_TTL` and only have `if token_ttl` or `auto_refresh` directly as an attribute in the class. Maybe im overthinking, but thats the current status. I will ponder this some more.

johnraz: > If we reference AUTO_REFRESH value in the authenticate/_credentials() methods it needs to be accessible on every view

I think we could solve this with something like:
```
if getattr(self, AUTO_REFRESH, knox_settings.AUTO_REFRESH):
    self.renew_token(auth_token)
```

instead of the current:
https://github.com/James1345/django-rest-knox/blob/c1972fcddf86179680522d6af1b463b02c49b00c/knox/auth.py#L77-L78

That way the order of precedence would be:
Class attribute's value first and next settings value, which is what we want.

sphrak: Thats a very nice solution, I like it. This PR does add some complexity to testing, I think I will have to mock some parts which tests the TokenAuthentication class directly because in that context there wont be any view. Is it okay if I mock it? no tests currently employs mocking.

johnraz: Cool! Glad you made it work 👍

knox/views.py
```diff
@@ -79,3 +79,30 @@ def post(self, request, format=None):
...
+
+    authentication_classes = (TokenAuthentication,)
+    permission_classes = (IsAuthenticated,)
+    AUTO_REFRESH = False
+
+    def post(self, request, format=None):
+
+        if not knox_settings.ENABLE_REFRESH_ENDPOINT:
+            return Response(
+                status=status.HTTP_403_FORBIDDEN
```
johnraz: Maybe adding a message here would be nice:

```
return Response(
    data={"error": "This view is not enabled"},
    status=status.HTTP_403_FORBIDDEN,
)
```
Not sure about the data returned's format though...

knox/views.py
```diff
@@ -79,3 +79,30 @@ def post(self, request, format=None):
...
+            )
+
+        if knox_settings.TOKEN_TTL is None:
+            raise ValueError(
+                'Value of \'TOKEN_TTL\' cannot be \'None\' in this context.'
+            )
```
johnraz: This will trigger a 500, I think we should make a note about the error code in the doc.

Also I would rather make this check at application loading time so it blows when you try to run the server and not when someone tries to hit the view... I'm not sure how to achieve that easily.

I'm not going to block if it stays as it is with a note in the doc though.

tests/tests.py
```diff
@@ -366,3 +374,107 @@ def test_expiry_is_present(self):
...
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(len(response.data), 1)
+        self.assertIn(
+            'expiry',
+            response.data
+        )
+        expected_expiry = original_time + ttl + timedelta(hours=5)
+        self.assertEqual(
+            make_naive(response.data['expiry']),
```
johnraz: Can't we build a timezone aware date to begin with instead of making this one naive?

tests/tests.py
```diff
@@ -366,3 +374,107 @@ def test_expiry_is_present(self):
...
+            self.assertEqual(response.status_code, 200)
+
+        # token does not work after new expiry:
+        new_expiry = AuthToken.objects.get().expiry
+        with freeze_time(new_expiry + timedelta(seconds=1)):
+            response = self.client.get(root_url, {}, format='json')
+            self.assertEqual(response.status_code, 401)
+
+    def test_endpoint_not_enabled_403_forbidden(self):
+
```
johnraz: Add a comment that by default the settings is false and the endpoint is thus disabled.

tests/tests.py
```diff
@@ -366,3 +374,107 @@ def test_expiry_is_present(self):
...
+        self.client.credentials(HTTP_AUTHORIZATION=('Token %s' % token))
+        five_hours_later = original_time + timedelta(hours=5)
+
+        url = reverse('knox_refresh')
+        with freeze_time(five_hours_later):
+            response = self.client.post(
+                url,
+                {},
+                format='json'
+            )
```
johnraz: We don't need freezetime magic in this test, just calling the view with a regular token is enough.

tests/tests.py
```diff
@@ -366,3 +374,107 @@ def test_expiry_is_present(self):
...
+
+        with override_settings(REST_KNOX=token_ttl_is_none_knox):
+            reload_module(views)
+            url = reverse('knox_refresh')
+            with freeze_time(five_hours_later):
+                with self.assertRaises(ValueError):
+                    self.client.post(
+                        url,
+                        {},
+                        format='json'
```
johnraz: Same as above, we don't need the freezetime magic.

knox/auth.py
```diff
@@ -37,6 +37,8 @@ class TokenAuthentication(BaseAuthentication):
     model = AuthToken
 
     def authenticate(self, request):
+
+        view = getattr(request, 'parser_context', None)
```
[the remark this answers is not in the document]
sphrak: Instructions a tad unclear, would you like me to rewrite those tests?

knox/views.py
sphrak: The error is documented here: https://github.com/James1345/django-rest-knox/pull/158/files#diff-1819b1daaccb3d358620ade9c67e9118R66

But I will add a note saying it will cause a HTTP 500 if improperly configured. But I agree, I'd rather have it fail immediatly at runtime rather than when a user tries to refresh, but am unaware of how.

knox/views.py
sphrak: I am not sure I agree -- I dont think configuration specific messages for the library should be exposed to the users here, no value to enduser imo. Are you sure?

knox/auth.py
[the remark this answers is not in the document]
johnraz: I think the tests shouldn't be failing yes and I guess this is because we use request factory with too few arguments.

knox/urls.py
```diff
@@ -4,6 +4,7 @@
 
 urlpatterns = [
     url(r'login/', views.LoginView.as_view(), name='knox_login'),
-    url(r'logout/', views.LogoutView.as_view(), name='knox_logout'),
-    url(r'logoutall/', views.LogoutAllView.as_view(), name='knox_logoutall'),
+    url(r'logout/$', views.LogoutView.as_view(), name='knox_logout'),
+    url(r'logout/all/$', views.LogoutAllView.as_view(), name='knox_logout_all'),
+    url(r'refresh/$', views.TokenRefreshView.as_view(), name='knox_refresh')
```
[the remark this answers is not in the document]
johnraz: Makes sense 👍

knox/views.py
johnraz: I do agree that exposing config related stuff is not a good practice.
Also if I'd get a 403 from a view I'd start looking around first into my authentication / permission system not at my config...
I was trying to reduce a bit of the indirection here - not gonna block or anything - just wanted to make it easier to reason about.

knox/views.py
johnraz: Works for me, we can investigate settings checking at runtime later on.

johnraz:
@sphrak to fix the current conflict, just remove docs/changes 😉

belugame:
@sphrak @johnraz  Vetos against releasing current master? I want to get your work out of the door, as things seem to be stuck at the moment. We will never run out of version numbers :)

sphrak:
@belugame I say go for it, release it. I unfortunatly won't have time for a couple more months to work on the stuck #156 issue right now. I will make a comment at least if anyone else want to work on it in the meantime.

johnraz:
Go ahead yes 👍 😊

belugame:
4.0.0 is on pypi
https://pypi.org/project/django-rest-knox/4.0.0/

PR #162, opened 2019-01-17
Change FALSE to False in code snippet

DougBeney:
I might be misunderstanding something but I think this should be False instead of FALSE.


johnraz:
Lgtm

belugame:
thank you @DougBeney  :)

PR #173, opened 2019-03-26
Drop require on pyOpenSSL and explicitly require cryptography

joonas:
It seems that the requirement for having `pyOpenSSL` was [dropped quite a while back](https://github.com/James1345/django-rest-knox/pull/71) (and [documented as such](https://github.com/James1345/django-rest-knox/blob/master/CHANGELOG.md#303)), so the only purpose `pyOpenSSL` now serves is to bring in `cryptography`, which would perhaps be better called out as a dependency directly.

I'm submitting this, because I felt that requiring `pyOpenSSL` was a little bit confusing when the only purpose it now serves is bringing in `cryptography`.


dk-WZFinTech:
I cheers idea of this - when we can expect this to be include and release?

PR #176, opened 2019-04-10
Add Django 2.2

jayvdb:


jayvdb:
We need to use xenial otherwise Django complains "SQLite 3.8.3 or later is required" https://travis-ci.org/jayvdb/django-rest-knox/builds/518272180

belugame:
Thank you @jayvdb!

belugame:
released as 4.0.1

Issue #171, opened 2019-03-14
Fully customisable LoginView response


josh-stableprice:
Currently, you require a user to override the post method in the event we need to change the format of the full response from LoginView. 

I suggest instead taking a parameter for the full response serializer `KNOX_LOGIN_RESPONSE_SERIALIZER`, with a method for us to override to provide context, with the token included. `def get_response_context(self, token)`

johnraz:
Hi @josh-stableprice,

I'm very late on the ball here, sorry for not giving you some feedback.

I think you raise a very valid point and I've been thinking about this for quite a while as well.

The way I would approach this would be to provide a utility method that is passed the `token`, `AuthToken` instance, and request.

Basically this all code https://github.com/James1345/django-rest-knox/blob/05f218f1922999d1be76753076cf8af78f134e02/knox/views.py#L44-L55

Should be moved to a method that would look like 
```
def get_post_response(self, request, token, auth_token):
    UserSerializer = self.get_user_serializer_class() 
  
    data = { 
        'expiry': auth_token.expiry, 
        'token': token 
    } 
    if UserSerializer is not None: 
        data["user"] = UserSerializer( 
            request.user, 
            context=self.get_context() 
        ).data 
    return Response(data)
```

One could then simply override `get_post_response` to handle the returned payload as desired.

/cc @belugame @sphrak @James1345

johnraz:
Reading my own message I think having a `get_post_response_data` method that returns the data dict would probably be better to override than a method that returns a full blown response.

johnraz:
I believe we can assume this has been covered by #180 - please re-open if something is missing.

PR #179, opened 2019-04-27
Add the possibility to configure a serializer for expiry in LoginView

### LINKED ISSUE #178: Configure expiry serilializer
With the new release `4.0.x` we now return the `expiry` date which is great, but it cannot be formatted the way I want it to be without overloading the [post](https://github.com/James1345/django-rest-knox/blob/master/knox/views.py#L30) method of `LoginView`.

I would like to be able to return the `expiry` directly as a Unix Timestamp instead.
An easy way to achieve this goal could be to add a configurable expiry serializer, we could then import from `knox_settings`. (basically in the same fashion it has been done for `UserSerializer`.

I would be willing to provide such a patch if you guys agree with the idea.

loicgasser:
Fixes #178 

This is what I had in mind.

- We keep the default current behaviour for expiry
- We define a new default `Serializer` class for exipiry in knox with a simple `DateTimeField`, which will take the default datetime formatting of the current project.
- We allow the user to define his own serializer for expiry via `ExpirySerializer` in the settings.

If you agree with these changes, I'll go ahead and adapt the doc.


loicgasser:
ping @johnraz what do you think?

loicgasser:
No problem, thank you for taking the time to review.

> get_expiry_serializer_class() should return the default serializer and in this context, ExpirySerializer shouldn't be None, ever.

Yes I think it's better not to either. I didn't know if you wanted to preserve the current behavior where the `datetime` field is not serialized explicitly in knox.

> I'm wondering if the Serializer isn't a bit too much and introduce indirection and unforeseen use case that would make maintaining backward compatibility harder.

Yes I was actually wondering the same, but defining only a field seemed a bit unnatural. I would propose `EXPIRY_FIELD` instead, because of `expiry` refers to the response key, and `field` the expected DRF class, same concept as `USER_SERIALIZER`, but I am fine with `EXPIRY_FORMAT` or `DATE_FORMAT` if you prefer it.

johnraz:
I'd argue that "expiry" should always be a datetime field, I don't quite see the need for having another field type - do you see a use case where you'd want something else than a datetime ?

johnraz:
Also, @loicgasser give a look at https://github.com/James1345/django-rest-knox/issues/171#issuecomment-490994892.
This could be a more generic approach to offer a way to customize the returned payload without cluttering the library with settings.

loicgasser:
> I'd argue that "expiry" should always be a datetime field

Yes, in my case I am returning an integer. I return all dates as UNIX timestamps. If you use `momentjs` on the front-end to handle dates, that's probably what you would want to have. I think it's a common use case for SPAs.

> Also, @loicgasser give a look at #171 (comment).
This could be a more generic approach to offer a way to customize the returned payload without cluttering the library with settings.

yes, both approaches are not mutually exclusive though. As a user of `knox`, I'd rather configure a field in the settings.

Let me know what is your decision on that and I'll adapt the code accordingly.

johnraz:
> Yes, in my case I am returning an integer. I return all dates as UNIX timestamps. If you use momentjs on the front-end to handle dates, that's probably what you would want to have. I think it's a common use case for SPAs.

I understand the need but I'm not so sure it's a common use case, it's really hard to tell.
You could also just use the format "%s" to represent the date, return a string in the payload and then deal with the type conversion client side.

> yes, both approaches are not mutually exclusive though. 

I totally agree

> As a user of knox, I'd rather configure a field in the settings.

This would indeed be very flexible but I'm not sure we could instantiate a field in the settings like that: 

* is DRF loaded yet? 
* is a field instantiated globally, safe to reuse between calls?

If the above works, let's go for it, if not let's fall back to the format string.

I'd also like input from other collaborators of the project (@belugame , @sphrak , @James1345 )

belugame:
@johnraz  I like your idea of having it solved with the same concept as e.g. Django's SHORT_DATETIME_FORMAT. Keep it small and simple. :+1:

loicgasser:
Here is my proposal.

Per default the new setting `EXPIRY_DATETIME_FORMAT` is formatted using the [DATETIME_FORMAT of Django Rest Framework](https://www.django-rest-framework.org/api-guide/settings/#datetime_format) with a `DateTimeField`. It doesn't solve my use case but I can understand you don't want to support this out of the box.

Additionally, we implement @johnraz proposal where we separate the formatting of the response by adding `get_post_response` as an additional method on the login view which can be easily overridden to return any custom response to support virtually all use cases.

johnraz:
@loicgasser let's do that and let's open a new PR for #171 (I'll add some more on the conversation there because I thought about some changes).

PR #180, opened 2019-05-24
Add setting for expiry format, add helper methods in login view

### LINKED ISSUE #171: Fully customisable LoginView response

### LINKED ISSUE #178: Configure expiry serilializer

loicgasser:
Replaces #179 
Fixes #171 
Fixes #178 

- Now expiry format defaults to whatever is used Django REST framework
- This behavior can be overriden via `EXPIRY_DATETIME_FORMAT` setting
- fully customizable expiry format via `format_expiry_datetime`
- fully customizable response payload via `get_post_response_data`

### DIFF
--- a/docs/settings.md
+++ b/docs/settings.md
@@ -9,6 +9,7 @@ Example `settings.py`
 #...snip...
 # These are the default values if none are set
 from datetime import timedelta
+from rest_framework.settings import api_settings
 REST_KNOX = {
   'SECURE_HASH_ALGORITHM': 'cryptography.hazmat.primitives.hashes.SHA512',
   'AUTH_TOKEN_CHARACTER_LENGTH': 64,
@@ -16,6 +17,7 @@ REST_KNOX = {
   'USER_SERIALIZER': 'knox.serializers.UserSerializer',
   'TOKEN_LIMIT_PER_USER': None,
   'AUTO_REFRESH': False,
+  'EXPIRY_DATETIME_FORMAT': api_settings.DATETME_FORMAT,
 }
 #...snip...
 ```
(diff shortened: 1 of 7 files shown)


loicgasser:
Thanks, no problem. It was a pleasure.

loicgasser:
@belugame any chance to have a second pair of eyes on this and maybe a version bump?

docs/settings.md
```diff
@@ -74,6 +76,14 @@ in the database.
 ## AUTH_HEADER_PREFIX
 This is the Authorization header value prefix. The default is `Token`
 
+## EXPIRY_DATETIME_FORMAT
+This is the expiry datetime format returned in the login view. The default is the
+[DATETIME_FORMAT][DATETIME_FORMAT] of Django REST framework. May be any of `None`, `iso-8601`
+or a Python [strftime format][strftime format] string.
+
+[DATETIME_FORMAT]: https://www.django-rest-framework.org/api-guide/settings/#date-and-time-formatting
+[strftime format]: https://docs.python.org/3/library/time.html#time.strftime
```
belugame: :+1:  very nice!

Issue #159, opened 2019-01-16
Make User and AuthToken model pluggable


pupeno:
[```User``` on line 24](https://github.com/James1345/django-rest-knox/blob/master/knox/auth.py#L24) is never used and [```model``` on line 39](https://github.com/James1345/django-rest-knox/blob/master/knox/auth.py#L39) is never used.

If they were used the way the seem to be intended and user was inside the class, this library would be much more flexible and usable in multi-user-model situations.

If this is something you are happy to accept, I can create the PR.

johnraz:
Hello @pupeno, thanks for reaching out.

I see a couple of things that could be improved and match with what DJRF does in the [`TokenAuthentication` class](https://github.com/encode/django-rest-framework/blob/master/rest_framework/authentication.py#L147):

- Have  a configurable `authenticate_header` keyword instead of always using `Token`
- Have a configurable token model instead of being forced to use `AuthToken`

Im' not sure to understand what you mean by "multi-user-model situations" and how having the the User within the class would be of any use?
I don't think we need the user model at all, and this code can probably be removed.

pupeno:
My application has three different classes (and tables) that represent users, one is the normal Django users and the other two are domain specific. Having the user model defined inside the class would allow to do something like this:

```python
class BuyerAuthToken(AuthToken):
    user_model = Buyer

class SellerAuthToken(AuthToken):
    user_model = Seller
```

johnraz:
Ok i see the usecase.
I personally think this is quite niche to use multiple user model.

I need to see a code proposal to decide if the complexity is worth the feature.

I would split it in 2 PRs one for the authentication class and one for the model.

pupeno:
I think the reason it's niche is because Django pushes people in the opposite direction. If it was possible and easier (like it is on other frameworks), more people would do it.

At my company we are now using this library so it'd be great if we can integrate our changes and maintain a single library instead of two separate ones. We are on crunch mode right now, but once that ends (and also we learned more about what works and what doesn't in this auth situation), I'll port the changes back into PRs.

johnraz:
I suggest we don't end up discussing Django's design here, because it will be a rabbit hole.
Let's focus on keeping this library maintainable and in sync with how Django is mainly used.

So as I said, let's decide when we see the code and its complexity.

Good luck with the crunch mode 👍

pupeno:
Done renaming.

[...]

PR #184, opened 2019-07-01
Make AuthToken model pluggable [WIP]

kalekseev:
Partially fixes https://github.com/James1345/django-rest-knox/issues/159.

This is a proof of concept that the feature is not that complex and can be added with minimal breaking changes. Would like to receive comments on that before investing more time.


johnraz:
Hi! Thanks a lot for this. The history seems to be a bit messed up. Could you rebase and make sure only your commits appear. I’ll try to follow up with a review ASAP once that’s done.

kalekseev:
@johnraz done

kalekseev:
it seems that develop branch doesn't contain all the master commits so pr includes a couple of comments from master

johnraz:
You should base your work on develop, not master.

knox/auth.py
```diff
@@ -34,7 +34,9 @@ class TokenAuthentication(BaseAuthentication):
     - `request.user` will be a django `User` instance
     - `request.auth` will be an `AuthToken` instance
     '''
-    model = AuthToken
+
+    def get_model(self):
+        return get_token_model()
```
johnraz: I would prefer to have this declared on the top of the file, in the same fashion the `User` model is declared.
What would be the use case to override this method?

knox/auth.py
```diff
@@ -62,7 +64,8 @@ def authenticate_credentials(self, token):
         '''
         msg = _('Invalid token.')
         token = token.decode("utf-8")
-        for auth_token in AuthToken.objects.filter(
+        model = self.get_model()
+        for auth_token in model.objects.filter(
```
johnraz: This implies that the new "AuthToken" model should have a field `token_key`, I'm wondering if we should not make this more flexible and have a `get_tokens` method has another way to fetch back the token.

knox/models.py
```diff
@@ -19,7 +19,7 @@ def create(self, user, expiry=knox_settings.TOKEN_TTL):
 
         instance = super(AuthTokenManager, self).create(
             token_key=token[:CONSTANTS.TOKEN_KEY_LENGTH], digest=digest,
-            salt=salt, user=user, expiry=expiry)
+            salt=salt, user=user, expiry=expiry, **kwargs)
```
johnraz: Can you explain this change?

knox/models.py
```diff
@@ -38,5 +38,10 @@ class AuthToken(models.Model):
     created = models.DateTimeField(auto_now_add=True)
     expiry = models.DateTimeField(null=True, blank=True)
 
+    class Meta:
+        # Work around for a bug in Django:
+        # https://code.djangoproject.com/ticket/19422
+        abstract = 'knox' not in settings.INSTALLED_APPS
```
johnraz: Reading the ticket, this seems to only apply to pre "app loading" times in Django. I don't think we care.

knox/views.py
```diff
@@ -60,7 +60,7 @@ def post(self, request, format=None):
                     status=status.HTTP_403_FORBIDDEN
                 )
         token_ttl = self.get_token_ttl()
-        instance, token = AuthToken.objects.create(request.user, token_ttl)
+        instance, token = get_token_model().objects.create(request.user, token_ttl)
```
johnraz: As per my comment above, this still implies some constraint on how the custom AuthToken model would work as it needs to comply to what happens in the create method and also reuse the manager from the default model.

knox/models.py
kalekseev: Basically the problem is following, if end user want to inherit from knox.models.AuthToken then he will get `RuntimeError: Model class knox.models.AuthToken doesn't declare an explicit app_label and isn't in an application in INSTALLED_APPS.`

knox/auth.py
johnraz: Yes sure but still I was challenging the fact that we could make this even more flexible in the case someone needs it. If the point is to make the model pluggable, I would probably go for something fully pluggable unless it's proven too complex.

knox/models.py
johnraz: I still have the feeling that this is not relevant anymore, did you witness the problem on a recent / supported version of Django?

knox/views.py
johnraz: That or we could simply have a `create_token(request)` method.

knox/models.py
johnraz: Ok but then you need to allow those params to be passed everywhere create is used in the code and this is not the case right now.

kalekseev:
@alex-yobota sorry but my original plan is different from what maintainers want:

> Yes sure but still I was challenging the fact that we could make this even more flexible in the case someone needs it. If the point is to make the model pluggable, I would probably go for something fully pluggable unless it's proven too complex.

It's quite reasonable decision to make it fully customizable but I don't have resources to do that now. If I will have new projects with similar requirements in near future I will implement this.

alex-yobota:
> @alex-yobota sorry but my original plan is different from what maintainers want:
> 
> > Yes sure but still I was challenging the fact that we could make this even more flexible in the case someone needs it. If the point is to make the model pluggable, I would probably go for something fully pluggable unless it's proven too complex.
> 
> It's quite reasonable decision to make it fully customizable but I don't have resources to do that now. If I will have new projects with similar requirements in near future I will implement this.

Cool, thanks for trying anyway :)

knox/models.py
yotamgod: I'm not sure this is what you mean, but I can confirm that on Django 2.2 this (or the lack of this) causes issues making knox less plug-able.
For example, when inheriting from AuthToken and adding field to the model, creating a multi-column constraint (unique_together) is not possible (there are two tables).

kalekseev: yes, that the same issue that DRF auth token had, so I just copied their solution here

johnraz:
As this PR is quite old and it was decided to make this more pluggable I'm going to close for now.

Issue #190, opened 2019-09-04
Auto refresh / Auto renew per token and not as global option


MoisesRT:
Hello,
I encountered a need to do Auto refresh / renew per token and not as global option(now set with `AUTO_REFRESH`).

I already have an idea how to do implement it myself(with PR of-course),
I just wanted to know if anyone else find this useful

MoisesRT:
https://github.com/James1345/django-rest-knox/pull/192

PR #192, opened 2019-09-15
Auto refresh per token

MoisesRT:
After I created the issue: https://github.com/James1345/django-rest-knox/issues/190
I started to work on a solution for it because we need it in our Django project,
so I did it in a way that the current behavior won't be affected.


belugame:
@MoisesRT  could you resolve the migration file name conflict and i will merge and release :) thank you

mohammadreza76:
hi
for using refresh token,now i set auto_refresh true and MIN_REFRESH_INTERVAL equal 10 and i set token_ttl in settings.my question is after these works what should i do?and what front should do?

PR #200, opened 2020-02-04
Update serializers.py

lijoev:
I need to get the user id also along with the username field. So added 'id' in the field list


johnraz:
I suggest you roll your own serializer. Exposing internal ids like this is not something everyone would want and actually I would recommend using some kind of uuid instead. Thanks for your time anyway.

PR #202, opened 2020-02-08
Raise AuthenticationFailed for one all one word headers, not just "Token"

khink:
The current code in authenticate() seems to try to raise
AuthenticationFailed if the authentication header is just one word:

        if not auth or auth[0].lower() != prefix.lower():
            return None
        if len(auth) == 1:
            msg = _('Invalid token header. No credentials provided.')
            raise exceptions.AuthenticationFailed(msg)

However, if `auth` has just one item, `auth[0].lower()` is not likely
to equal the prefix. So probably we'll just return None.

This commit tries to fix that by testing
`auth[0].lower() != prefix.lower()` only after we've checked that we
received the correct number of words in the authentication header.

### DIFF
--- a/knox/auth.py
+++ b/knox/auth.py
@@ -37,7 +37,10 @@ class TokenAuthentication(BaseAuthentication):
         auth = get_authorization_header(request).split()
         prefix = knox_settings.AUTH_HEADER_PREFIX.encode()
 
-        if not auth or auth[0].lower() != prefix.lower():
+        if not auth:
+            return None
+        if auth[0].lower() != prefix.lower():
+            # Authorization header is possibly for another backend
             return None
         if len(auth) == 1:
             msg = _('Invalid token header. No credentials provided.')
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -5,6 +5,7 @@ from django.contrib.auth import get_user_model
 from django.test import override_settings
 from django.utils.six.moves import reload_module
 from freezegun import freeze_time
+from rest_framework.exceptions import AuthenticationFailed
 from rest_framework.serializers import DateTimeField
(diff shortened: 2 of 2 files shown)


johnraz:
I dont understand the rationale behind thinking that the prefix would not match ever ?

The header could very well be `Prefix` alone and if so, we can assume an empty token has been provided (with a good prefix) and raise properly.

The goal is to only report that a bad token has been used when we know that the prefix is correct. 

I dont think we need the proposed change.

I’m interested in the tests you added though, would you mind submitting only those?

Feel free to keep explaining if I miss something 😊

khink:
Hi, thanks for explaining your point of view. Indeed the situation you mention is the one situation  where the current code enters the part where `if len(auth) == 1:` evaluates to true. 

But if a client erroneously submits one of these:

- `Tokensometoken` in error, so without the space
- only a token key (`sometoken`), without the "Token " prefix
- some other header that is only one word

, then `auth[0].lower() != prefix.lower()` would also evaluate to True, and `None` would be returned. I would argue that in that case `Invalid token header.` would be the better error message, or maybe even something like "Received only one word in auth header". We might even check if that word equals "Token", to give a more precise message (like "Token key missing from auth header").

I guess you know more about the rationale behind this than i do, so if this is intended behavior i'll just leave the code unchanged and change the new tests so they cover all code paths.

khink:
I now understand that some other authentication backend might want to use the `HTTP_AUTHORIZATION` header, and that is why we want to return `None` instead of  raising an error. I'll adapt tests accordingly.

johnraz:
Thanks 🙏

khink:
Thank *you* for your quick and constructive feedback!

Issue #194, opened 2019-10-17
auth.py calls get_user_model() which breaks Django when using a custom user model.


brumiser1550:
I cannot get Django to even consider running while the auth.py file calls get_user_model(). 

If I take it out everything works. It looks like this is not even used so not sure why it is even needed in this file. 

```
File "/var/www/pycharmprojects/feedback_project/venv/lib/python3.6/site-packages/django/contrib/auth/__init__.py", line 170, in get_user_model
    "AUTH_USER_MODEL refers to model '%s' that has not been installed" % settings.AUTH_USER_MODEL
django.core.exceptions.ImproperlyConfigured: AUTH_USER_MODEL refers to model 'users.User' that has not been installed
```

alexhalford:
Have you installed your users app?

johnraz:
Old and lacks details, closing.

PR #203, opened 2020-02-08
Remove unused User definition

khink:
I was triggered to look into this by https://github.com/James1345/django-rest-knox/issues/194. It looks like the reporter of that issue has had some kind of misconfiguration, and that issue can be closed. 

But it does look like the definition of `User` is superfluous here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ matrix:
   include:
     - python: "2.7"
       env: TOX_ENVS=py27-django111
-    - python: "3.4"
-      env: TOX_ENVS=py34-django111,py34-django20
     - python: "3.5"
       env: TOX_ENVS=py35-django111,py35-django20,py35-django20,py35-django21
     - python: "3.6"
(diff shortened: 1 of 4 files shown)


zshongyi:
I have the same problem with #194.
If i set REST_FRAMEWORK['DEFAULT_AUTHENTICATION_CLASSES'] as ('knox.auth.TokenAuthentication',)
django start up failed, and report :
 AUTH_USER_MODEL refers to model 'XXX' that has not been installed.

It look like django install apps failed which include AUTH_USER_MODEL.

import AUTH_USER_MODEL need import api_settings.DEFAULT_AUTHENTICATION_CLASSES (site-packages\rest_framework\schemas\__init__.py", line 33).
however knox.auth.TokenAuthentication module need import AUTH_USER_MODEL(\site-packages\knox\auth.py", line 22, User = get_user_model()）.
It's a dead circle
User never be used in this module,why not remove this line?

PR #204, opened 2020-02-08
Stop testing with Python 3.4

khink:
Python 3.4 is end-of-life (https://www.python.org/downloads/release/python-340/), we can stop supporting it.

Some builds (https://travis-ci.org/James1345/django-rest-knox/jobs/647661512, https://travis-ci.org/James1345/django-rest-knox/jobs/647660925) are now failing because `PyYAML requires Python '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*' but the running Python is 3.4.8`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ matrix:
   include:
     - python: "2.7"
       env: TOX_ENVS=py27-django111
-    - python: "3.4"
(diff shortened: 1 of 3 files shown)


johnraz:
Thanks a lot !

PR #205, opened 2020-02-13
Add label as optional field to identify a token

ciur:
Hi! 

would it be possible to include optional label field to the model? I use knox auth in papermerge project and I find it confusing to identify tokens by their digest. From user's point of view, digests and tokens look similar.

[Here is the use case.](https://papermerge.readthedocs.io/en/latest/rest_api.html#use-the-token)

Note that tokens are **not** created in login view (I dont use knox' login view at all).


ciur:
I don't understand what/why exactly CI failed :(

johnraz:
The CI fails because you didn't add the model migrations.

To be honest I don't find this to be generic enough to be included in the project, you could easily attach a metadata object to the model and customise things as you wish.

PR #208, opened 2020-03-20
Update settings

leonardfreitas:
missing an "I" in "api_settings.DATETME_FORMAT"


leonardfreitas:
I saw afterwards that this had already been sent.

PR #210, opened 2020-04-28
Use six directly in order to work with django 3.x

scarabeusiv:


asencis:
Are there any updates on this issue?

PR #214, opened 2020-06-09
fix django warnings - ugettext_lazy is deprecated

foogunlana:
```
RemovedInDjango40Warning: django.utils.translation.ugettext_lazy() is deprecated in favor of django.utils.translation.gettext_lazy().
```

See: https://docs.djangoproject.com/en/3.0/_modules/django/utils/translation/

PR #225, opened 2020-09-26
add new python and django to test

Rotzbua:
update travis ci, setup.py, tox
add django 3.0 3.1
add python 3.8
remove django 1/2.0/2.1 
remove python 2.7


tolomea:
Is this likely to be merged? and if so when is it likely to release?

Rotzbua:
An additional note:
with pr #214 there will be problems with unicode translations under python 2.7 , so official support should be dropped

Issue #199, opened 2020-01-23
Remove cryptography dependency?


tasn:
sha512 is already [in hashlib](https://docs.python.org/3.8/library/hashlib.html) so the cryptography dependency is not really needed. This is especially important since the cryptography dep can be a hassle to setup in some cases (requires a compiler, headers and openssl, as per django-rest-knox docs).

It's a very easy change to make and won't break existing applications as long as they haven't manually set the hash setting and override it to something from cryptography. If they have: this can be a breaking change for a future version.

Rotzbua:
I agree. Seems it was already modernized in #173, but modern python3 does not require this dependency anymore. I created a pr.

PR #230, opened 2020-10-13
Use hashlib

Rotzbua:
## Purpose

Remove dependency to extra library which maybe require compiling.
Internal hashlib library has the same function. For default sha512 it does not require openssl. If openssl is available it is used for better performance.

## Reference

Issue #199

## Notice

~~This pr includes #225 because it does not work on old python 2.7.~~

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -2,27 +2,8 @@
 
 ## Requirements
 
-Knox depends on `cryptography` to provide bindings to `OpenSSL` for token generation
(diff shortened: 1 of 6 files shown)


johnraz:
@Rotzbua I merged some of you other PR and this sadly created some conflicts, if you don't mind looking into that :-)

Rotzbua:
Hi johnraz, I rebased my pr.
The change will brake backwardscompability if somebody uses a custom hashing function from old cryptography library.

PR #238, opened 2020-11-15
fix deprecation of djangos url() #237

### LINKED ISSUE #237: Djangos url() is deprecated
## Warning

> RemovedInDjango40Warning: django.conf.urls.url() is deprecated in favor of django.urls.re_path().

## Fix

Replace by `re_path` or `path`.

## Why?

Prepare for upcoming django.

Rotzbua:
## Description
Just maintenance, should not have a real impact.

## Issue

Fixes #237

### DIFF
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -65,9 +65,9 @@ from knox import views as knox_views
 from yourapp.api.views import LoginView
 
 urlpatterns = [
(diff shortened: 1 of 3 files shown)


johnraz:
thx for the contrib

SyedAhkam:
The docs still haven't been updated to reflect `path` instead of `url`?

Source: https://james1345.github.io/django-rest-knox/auth/

PR #239, opened 2020-12-01
Update settings.md

nnamdei:


johnraz:
This is fixed in develop

PR #262, opened 2022-01-19
Add @smart_token_user decorator to get user object easier

xros:
I wrote a decorator.

With this,
in our app views.py
we can easily get user object by doing so,`@smart_token_user` will modify the request handler. We can have a `request.user` attr only once the token is valid. And all invalid attempts will be thrown out with HTTP 401 Unauthorized response.
Life can be easier with this decorator.

```python
from knox.models import smart_token_user

class CheckUserEmail(generics.RetrieveAPIView):
    @smart_token_user
    def get(self, request):
        return Response({
            "username": request.user.username,
            "email": request.user.email,
            "password": request.user.password,
        }, status=status.HTTP_200_OK)
```


giovannicimolin:
@xros Thank you for your contribution! :grin: 

I'm understanding the purpose of these changes. If you are using Know with Django REST Framework, DRF's internal mechanisms will already handle the authentication step and make the user object available through `request.user` without any extra code.

Can you explain a bit better the case where you ran into issues, so I can check if your use case is valid?
Let me know if you still have this issue.

Best!

giovannicimolin:
Closing this since it's been a while and I haven't heard back.

PR #268, opened 2022-06-17
fix(docs): add warning of migration to `4.2.0`

### LINKED ISSUE #266: Version 4.2.0 breaks use of tokens generated with older versions
#193, which is included in 4.2.0, removed the usage of the salt when generating the hash digest for API tokens. That means digests generated for a token will differ depending on whether `django-rest-knox<4.2.0` or `django-rest-knox==4.2.0` is used, as older versions use the salt and token to generate the hash digest, while 4.2.0 only uses the token without a salt.

As `django-rest-knox` stores the hash digest and not the token itself, authentication with tokens generated with `django-rest-knox<4.2.0` when using `django-rest-knox==4.2.0` doesn't work anymore, as the hash digest generated when creating the token was generated using token and salt as input, while the one generated when checking for authentication now only uses the token as input.

Please note that this is a serious breaking change from 4.1.0 to 4.2.0 which isn't mentioned anywhere and renders all previously issued tokens invalid!

Rotzbua:
Just a documentation addition according to and closes #266.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 4.2.0
 - compatibility with Python up to 3.10 and Django up to 4.0
 - integration with github CI instead of travis
-- Migration: "salt" field of model "AuthToken" is removed
+- Migration: "salt" field of model "AuthToken" is removed, WARNING: invalidates old tokens!
 
 ## 4.1.0
 


giovannicimolin (approved):
:+1: 

Thanks for the contribution!

PR #278, opened 2022-08-30
Add Django 4.1 to test 

Rotzbua:
Django 4.1 has been released!
https://docs.djangoproject.com/en/4.1/releases/4.1/


tox.ini
```diff
@@ -46,3 +48,4 @@ python =
     3.8: py38
     3.9: py39, isort, flake8
     3.10: py310
+    3.10: py311
```
joaonc, line 51: Should this be `3.11: py311` instead of `3.10` ?

Issue #220, opened 2020-08-21
Migrating from rest_framework.TokenAuthentication without logging users out


oaosman84:
Is there a recommended way to migrate to knox from DRF's TokenAuthentication—without logging users out?

If I swap in knox instead of DRF, then all users need to re-login (since their old token isn't valid). I can't seem to get both systems working simultaneously, because `knox.TokenAuthentication` raises an exception (instead of returning None) if the the token is invalid, which causes Django not to try subsequent authenticators.

Ideally, I'd like a way to try knox.TokenAuthentication, then if it fails but the token works in DRF, create a knox token for that user (then at some point in the future, completely remove DRF tokens).

joelawm:
I suppose if you're running say a redux store you could check the token in from the browser storage/store and see if in the DRF table. If it is remove it and send back a new token from the knox table. Then slowly phase that out until the DRF table is empty or a majority of active users have switched.

jonathan-golorry:
Something like this would work:

```
from django.db import migrations
from knox import crypto
from knox.settings import CONSTANTS


def convert_tokens(apps, schema_editor):
    Token = apps.get_model("authtoken", "token")
    AuthToken = apps.get_model("knox", "authtoken")
    db_alias = schema_editor.connection.alias

    for token in Token.objects.using(db_alias).all():
        digest = crypto.hash_token(token.key)
        # bypass custom manager
        AuthToken.objects.using(db_alias).bulk_create(
            [
                AuthToken(
                    digest=digest,
                    token_key=token.key[: CONSTANTS.TOKEN_KEY_LENGTH],
                    user_id=token.user_id,
                    expiry=None,  # or whatever logic you want
                )
            ]
        )
        # bypass auto_now_add restriction
        AuthToken.objects.using(db_alias).filter(digest=digest).update(
            created=token.created
        )

    Token.objects.using(db_alias).all().delete()


class Migration(migrations.Migration):

    dependencies = [
        ("knox", "0008_remove_authtoken_salt"),
    ]

    operations = [migrations.RunPython(convert_tokens)]
```

(keep in mind this is for the latest unreleased version of knox where the salt has been removed from AuthToken)

Remove "rest_framework.authentication.TokenAuthentication" from your views and default authentication, but keep both "rest_framework.authtoken" and "knox" in INSTALLED_APPS until the migration has run everywhere.

gawry:
@jonathan-golorry just created a PR with a "safe" version of your migration

Issue #215, opened 2020-06-26
Migrate Existing AuthTokens


steverecio:
I have users with live API integrations using the built-in DRF Auth Tokens. Is there a way to quietly migrate those existing tokens into knox tokens without disrupting service? I've installed the app locally but the existing DRF tokens didn't migrate automatically.

steverecio:
Perhaps its possible to support both DRF auth tokens and knox tokens simultaneously to give users a chance to switch over?

PR #295, opened 2023-01-11
Changes to allow a migration of DRF tokens (#220 and #215)

gawry:


giovannicimolin:
@gawry Thanks for the contribution! :grin: 

After reviewing your changes, I don't think the best place for this operation is a migration. The changes explicitly assume that people are using DRF's `authtoken` model, which isn't always the case.
With interest in keeping this library as agnostic to external libraries **as possible**, I'm inclined to reject these changes.

That said, there's some useful content in this PR that can become a nice piece of documentation guiding new users on how to perform the token migration on their applications. Are you interested in changing this PR to a documentation improvement?

gawry:
Hello @giovannicimolin, 

Thank you for taking the time to review my PR.

I understand the concern about keeping the library agnostic to external libraries. The intention behind my changes was to ensure that the operation would only be executed when the user is leveraging DRF's authtoken model. If they aren't using it, the migration would simply be a no-op and wouldn't have any impact.

By integrating this feature, we can provide a seamless experience for those who are using the authtoken model, without affecting those who aren't. The design has been crafted with the very purpose of ensuring that the library remains flexible and agnostic.

However, if you still believe that this shouldn't be part of the core migration, I'm more than happy to pivot this PR into something else. Perhaps a documentation piece or a `command`. This could guide users on how to handle token migration if they're using the authtoken model, as you suggested.

Looking forward to your feedback.

Best,
Gawry

PR #296, opened 2023-01-16
Enable customizing login/logout responses

c-w:
This pull request is a small change to enable customizing the response shape for the login and logout views by overriding a simple method.

This is useful in situations where for example the 204 responses in the logout views are undesirable (e.g. for clients that always assume that responses have a JSON body).

### DIFF
--- a/docs/views.md
+++ b/docs/views.md
@@ -67,12 +67,24 @@ It responds to Knox Token Authentication. On a successful request,
 the token used to authenticate is deleted from the
 system and can no longer be used to authenticate.
 
+By default, this endpoint returns a HTTP 204 response on a successful request. To
+customize this behavior, you can override the `get_post_response` method, for example
+to include a body in the logout response and/or to modify the status code:
+
+```python
+...snip...
+    def get_post_response(self, request):
+        return Response({"bye-bye": request.user.username}, status=200)
+...snip...
+```
+
 ## LogoutAllView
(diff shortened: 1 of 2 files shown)


AdamDonna:
This looks very reasonable to me 👍 
I think you're good to merge if I can get another +1.
@giovannicimolin does this look good to you?

giovannicimolin:
@c-w Thanks for the contribution! :grin: 

I'll be reviewing these changes in the next week or so.
I don't have a lot of bandwidth but I'll try to keep things moving.

CC @AdamDonna

c-w:
> This LGTM as well, maybe worth adding to the documentation though.

Where would you like me to document this @johnraz ?

johnraz:
i guess here: https://github.com/jazzband/django-rest-knox/blob/develop/docs/views.md

giovannicimolin (commented):
:+1: 

LGTM, changes look good and there's some interest in this being merged.

I think this is almost good to merge, but I agree with @johnraz that it needs some documentation. If you add those docs in, I'll make sure to review and move this along. :grin:

c-w:
> it needs some documentation

I added the docs in 43d983d. PTAL @giovannicimolin

giovannicimolin (approved):

giovannicimolin:
:+1: 

CI checks are green, we're good to go.
@c-w Thanks for your contribution. :grin:

PR #317, opened 2023-10-10
chore(migration): use f-string

Rotzbua:
Just some migrations to modern python.


giovannicimolin (approved):
:+1: 

Looks good. Codecov is complaining about coverage but it's for the `__str__`  method.

giovannicimolin:
@Rotzbua It's complained because of the changed lines is not covered by any tests. It's ok though.

Thanks for the PR! :grin:

PR #318, opened 2023-10-12
Count non expiring tokens when determining if the limit is reached

### LINKED ISSUE #280: Tokens with no maximum expiration time are not counted towards the maximum number of tokens allowed
**Description**
According to docs, setting the `TOKEN_TTL` to `None` will create tokens that never expire.

However, before token creation, when checking that the number active of tokens of a user < `token_limit_per_user`, tokens without expiry date (`expiry=None`) are not taken into account.

https://github.com/James1345/django-rest-knox/blob/3a1bc584f9691f4bc19d8a04a98c68c293be9ca6/knox/views.py#L65

**Expected behavior**

All active tokens, including the latter, would be expected to be taken into account.

**Possible solution**

I don't know if I'm misunderstanding something, but a change like the following would suffice.

```python
# knox/views.py
from django.db.models import Q

...

token = request.user.auth_token_set.filter(Q(expiry__gt=now) | Q(expiry__isnull=True)) 
```

morty:
Fixes #280

Thanks to @pablomm for the code in the ticket

### DIFF
--- a/knox/views.py
+++ b/knox/views.py
@@ -1,4 +1,5 @@
 from django.contrib.auth.signals import user_logged_in, user_logged_out
+from django.db.models import Q
 from django.utils import timezone
 from rest_framework import status
 from rest_framework.permissions import IsAuthenticated
@@ -62,7 +63,9 @@ class LoginView(APIView):
         token_limit_per_user = self.get_token_limit_per_user()
         if token_limit_per_user is not None:
             now = timezone.now()
-            token = request.user.auth_token_set.filter(expiry__gt=now)
+            token = request.user.auth_token_set.filter(
+                Q(expiry__gt=now) | Q(expiry__isnull=True)
+            )
             if token.count() >= token_limit_per_user:
                 return Response(
                     {"error": "Maximum amount of tokens allowed per user exceeded."},
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -339,8 +339,10 @@ class AuthTestCase(TestCase):
 
         with override_settings(REST_KNOX=token_user_limit_knox):
             reload(views)
-            for _ in range(10):
+            for _ in range(5):
                 AuthToken.objects.create(user=self.user)
+            for _ in range(5):
+                AuthToken.objects.create(user=self.user, expiry=None)
             url = reverse('knox_login')
             self.client.credentials(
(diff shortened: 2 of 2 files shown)


giovannicimolin:
@morty @pablomm Thanks for the contribution! That's a nice catch!

I have one concern with this change though. Once we start considering non-expiring tokens, it is possible that users end up getting locked out of their accounts after issuing a certain number of tokens without logging out properly.

We either need this behind settings flag or a way for new log-ins to remove the oldest unused token.
Do you think my concern makes sense?

pablomm:
HI @giovannicimolin and thanks to @morty for turning the fix into a pull request.

In my case, I use these non-expiring tokens for users that uses my platform through an API. They have access to a control panel (which they access with another authentication system) where they can create, view and delete their api tokens (knox tokens which may be non-expiring). So logging out does not make sense in my use case.

The problem you comment, wouldn't it also happen if someone creates too many tokens within the expiration time?

morty:
To get to this situation two settings (`TOKEN_TTL` and `TOKEN_LIMIT_PER_USER`) would have had to be changed from their default values. That's a deliberate choice. For `TOKEN_LIMIT_PER_USER` to be ignored without warning if `TOKEN_TTL`  is `None` seems dangerous. Maybe a warning in the documentation for `TOKEN_LIMIT_PER_USER` saying that when the limit is reached users won't be able to get another one until one of their tokens has timed out and if they are non-expiring then will be locked out until they can use another method to delete a token?

My client is using this with a very small number of end users where they could handle support requests to delete tokens if required. I could see this wouldn't be possible where there are a very large number of end users.

If `LogoutAllView` allowed other authentication methods like `LoginView` then the user could clean out their tokens with username and password. But it looks like this might have been explored before based on the warning in the docs for the `LogoutAllView`.

johnraz:
@morty @giovannicimolin I believe the change made here makes full sense, this is clearly a bug that needs to be fixed.

I would add a note to the documentation (and probably a comment in the code) that clearly explains the consequence of having dangling non-expiring tokens around.

Using non-expiring tokens should be also discouraged as it's a bad practice anyway.

foucdeg:
hello @giovannicimolin , has this been released?

PR #321, opened 2023-12-10
feat(ci): add current django and python to test

Rotzbua:
- Django 4.0 EOL since 01 Apr 2023.
- Django 4.1 EOL since 01 Dec 2023.
- 4.2 is new LTS.
- 5.0 is new current.

https://endoflife.date/django


dekkers (approved):

dekkers:
We should probably also remove Python 3.6 and 3.7 as they are also end of life.

PR #335, opened 2024-04-26
Avoid importing `django.test` in main code

angusholder:
`setting_changed` is actually from `django.core.signals`, and is re-exported from `django.test.signals`.

`django.test` takes 20-40ms to import, so not too much, but it's an easy win to avoid this import.

### DIFF
--- a/knox/settings.py
+++ b/knox/settings.py
@@ -1,7 +1,7 @@
 from datetime import timedelta
 
 from django.conf import settings
-from django.test.signals import setting_changed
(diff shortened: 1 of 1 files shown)


giovannicimolin:
@angusholder Thanks for your contribution! 

That's a nice catch! I can accept that right away.

giovannicimolin (approved):
:+1: 

Looks good! 
Tests are passing and this is a good and simple import fix.

PR #342, opened 2024-05-02
Prepare next release

johnraz:
Some things that need to be done before develop is ready:

- [x] Changelog should be updated
- [ ] Base on the changelog update a version should be promoted
- [x] A task to generate the documentation and publish the github page should be put in place
- [x] Access to the pypi registry should be gained

Keep this PR in draft until the above are done

@James1345 @giovannicimolin

PR #345, opened 2024-05-06
Remove EOL versions

calumy:
This PR removes the following end-of-life versions of [Python](https://devguide.python.org/versions/) and [Django](https://jefftriplett.com/django-release-cycle/): 
- Python 3.6 - EOL: 23rd December 2021
- Python 3.7 - EOL: 27th June 2023
- Django 3.2 - EOL:  April 2024 

In addition to removing the EOL versions, the [Django upgrade pre-commit hook](https://github.com/adamchainz/django-upgrade) has also been added, and fixes are included in this PR. These changes relate to:
- The use of `USE_L10N`, which was deprecated in Django 4.0: https://github.com/adamchainz/django-upgrade?tab=readme-ov-file#use_l10n
- The use of `re_path`, which has been rewritten using the `path()` syntax added in Django 2.0: https://github.com/adamchainz/django-upgrade?tab=readme-ov-file#urls

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       max-parallel: 5
       matrix:
-        python-version: ['3.6', '3.7', '3.8', '3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


johnraz (changes requested):
I was thinking we could merge this after the next release, like that we can issue a warning that this will be the last release to support all those older versions.

calumy:
Drafting until #342 is merged and a new version released.

johnraz (approved):
LGTM

giovannicimolin:
@calumy Thanks for your contribution! :)

I'll update the changelogs and documentation of the before the next release and be sure to add a note.

PR #346, opened 2024-05-08
Remove flake8 and isort from tox config

calumy:
Following on from the discussion in #344, this PR removes linting from the tox config to allow linting and testing to be run independently.


johnraz:
Thanks :)

PR #353, opened 2024-06-27
chore: remove outdated super() call

Rotzbua:
object not required anymore


giovannicimolin (approved):
:+1: 

Thanks for the cleanup!

PR #374, opened 2025-01-21
Use python3 syntax for super()

hrideshmg:
Came across this while trying to set the framework up. 
Not sure why but the given python2 syntax seems to error for my project citing `'super' object has no attribute 'post'`, migrating this to the python3 syntax seems to fix it.

### DIFF
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -90,7 +90,7 @@ class LoginView(KnoxLoginView):
         serializer.is_valid(raise_exception=True)
         user = serializer.validated_data['user']
         login(request, user)
-        return super(LoginView, self).post(request, format=None)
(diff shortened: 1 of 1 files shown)


johnraz:
LGTM, might be a good opportunity to adapt all the super calls at once?

hrideshmg:
> LGTM, might be a good opportunity to adapt all the super calls at once?

I don't believe there are any more super calls?

giovannicimolin (approved):
👍 

Looks good!

Since this is only a documentation fix I don't think it warrants a release. 

Thanks for the contirbution! 😁

hrideshmg:
Great, this is probably the fastest I've seen a PR get merged lol. Thanks 🙌

PR #378, opened 2025-02-15
Fix for n+1 query on AuthToken.user

naviens:
Fix for Potential n+1 query detected on `AuthToken.user`
Screenshot 2025-02-15 at 10 54 58 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 5.0.3
+- Fix for Potential n+1 query detected on AuthToken.user
+
 ## 5.0.2
 - Implement AUTO_REFRESH_MAX_TTL to limit total token lifetime when AUTO_REFRESH = True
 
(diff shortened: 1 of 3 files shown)


johnraz:
Hey!
Thanks for the contribution and good catch !

Would be nice if we could cover this with a test and more importantly can you add a note to the changelog as this might have side effects due to the added join ?

johnraz:
Thanks again!
@giovannicimolin as usual please have a look as well ☺️

naviens:
Hi Johnraz,

It's been a while! Looking forward to the PR being merged into the main branch.

johnraz:
As you can see from my previous message I’m still waiting for @giovannicimolin ’s feedback ;-)

giovannicimolin (approved):
👍 

Looks good! That's a very nice catch!

Apologies for the delay here. I'll see if I can get a few other PRs in before making another release.

